PR #7, opened 2015-12-11
Bug 1231697 - add a Firefox runner that links the add-on to aid in dev…

mikedeboer:
…elopment. r?Mardak

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,6 +218,10 @@ test: lint karma
 runserver: remove_old_config
 	$(REPO_BIN_DIR)/run-server.sh
 
+.PHONY: runfx
+runfx:
+	bin/runfx.js ${ARGS}
+
 frontend:
 	@echo "Not implemented yet."
 
--- a/bin/run-utils.js
+++ b/bin/run-utils.js
@@ -1,0 +1,192 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+/* eslint-env node */
+
+var Extend = require("lodash").extend;
+var Fs = require("fs-promise");
+var FxProfileFinder = require("firefox-profile/lib/profile_finder");
+var FxRunner = require("fx-runner/lib/run");
+var Path = require("path");
+var When = require("when");
+
+/**
+ * Profiles that do not include "/" are treated as profile names to be used by
+ * the Fx Profile Manager.
+ *
+ * @param {String} profile Name of the profile to test
+ * @return {Boolean}
+ */
+function isProfileName(profile) {
+  if (!profile) {
+    return false;
+  }
+  return !/[\\\/]/.test(profile);
+}
+
+/**
+ * Retrieve the valid path of a profile.
+ *
+ * @param  {String}  profile Name or path of the profile
+ * @return {Promise} Will be resolved with a valid absolute path pointing to the
+ *                   profile or rejected when the path does not exist or when the
+ *                   profile name can not be found.
+ */
+function getProfilePath(profile) {
+  if (isProfileName(profile)) {
+    var finder = new FxProfileFinder();
+    return When.promise(function(resolve, reject) {
+      finder.getPath(profile, function(err, path) {
+        if (err) {
+          reject(err);
+          return;
+        }
+
+        resolve(path);
+      });
+    });
+  }
+
+  return When.promise(function(resolve, reject) {
+    profile = Path.normalize(profile);
+    if (!Path.isAbsolute(profile)) {
+      profile = Path.join(process.cwd(), profile);
+    }
+    Fs.stat(profile).then(function(stat) {
+      if (!stat.isDirectory()) {
+        reject("Profile path '" + profile + "' is not a directory");
+        return;
+      }
+      resolve(profile);
+    }).catch(reject);
+  });
+}
+exports.getProfilePath = getProfilePath;
+
+// Environment variables that make sure that all messages and errors end up in
+// the process stdout or stderr streams.
+var FX_ENV = {
+  "XPCOM_DEBUG_BREAK": "stack",
+  "NS_TRACE_MALLOC_DISABLE_STACKS": "1"
+};
+
+/**
+ * Check a chunk of process output to check if it contains an error message.
+ *
+ * @param  {String} line
+ * @return {Boolean}
+ */
+function isErrorString(line) {
+  return /^\*{25}/.test(line) || /^\s*Message: [\D]*Error/.test(line);
+}
+
+var GARBAGE = [
+  /\[JavaScript Warning: "TypeError: "[\w\d]+" is read-only"\]/,
+  /JavaScript strict warning: /,
+  /\#\#\#\!\!\! \[Child\]\[DispatchAsyncMessage\]/,
+  /JavaScript strict warning: resource:\/\/\/modules\/sessionstore\/SessionStore\.jsm/,
+  /JavaScript strict warning: resource:\/\/gre\/components\/nsSearchService\.js/
+];
+
+/**
+ * Checks of the chunk of process output can be discarded. The patterns that are
+ * tested against can be found in the `GARBAGE` array above.
+ *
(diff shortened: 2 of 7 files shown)


.eslintignore
```diff
@@ -1,6 +1,7 @@
 # Libs we don't need to check
 node_modules
 **/vendor
+lib
```
Mardak: We can lint the run/util files.

Makefile
```diff
@@ -142,6 +142,10 @@ test: lint karma
 runserver: remove_old_config
 	$(REPO_BIN_DIR)/run-server.sh
 
+.PHONY: runfx
+runfx:
+	lib/run.js ${ARGS}
```
Mardak: We should probably do something similar to distserver where it calls node explicitly for server.js, so no need for the #! node. Similarly let's keep run/util.js in bin/ instead of lib/.

.eslintignore
Mardak: Sorry, forgot to mention except for 

  45:13  error  Don't use process.exit(); throw an error instead  no-process-exit
  75:5   error  Don't use process.exit(); throw an error instead  no-process-exit
  82:3   error  Don't use process.exit(); throw an error instead  no-process-exit

Mardak:
I ran this on a new profile and initially got:

1449865695391   addons.xpi-utils    WARN    Disabling foreign installed add-on loop@mozilla.org in app-profile

Is this because xpinstall.signatures.required needed to be set to false? If so, perhaps runfx should also default set that pref to false.

bin/runfx.js
```diff
@@ -0,0 +1,88 @@
...
+  var profile = Commander.profile || DEFAULT_PROFILE;
+
+  Util.getProfilePath(profile).then(function(profilePath) {
+    // Since we've got the profile path now, we can create the symlink in the
+    // profile directory IF it doesn't exist yet.
+    var addonTargetDir = Path.join(profilePath, "extensions", "loop@mozilla.org");
+    return ensureRemoved(addonTargetDir).then(function() {
+      // This should fail at a certain point, because we don't want the add-on
+      // directory to exist.
+      return Fs.symlink(addonSourceDir, addonTargetDir).then(function() {
```
Mardak: This fails if `extensions` directory doesn't exist within the profile directory:

```
{ [Error: ENOENT, symlink '/Users/Ed/loop/built/add-on']
  errno: 34,
  code: 'ENOENT',
  path: '/Users/Ed/loop/built/add-on' }
```

Mardak:
On a clean profile after creating `extensions` directory, about:addons shows a disabled add-on:

`Firefox Hello could not be verified for use in Nightly and has been disabled.`

`about:config` toggling `xpinstall.signatures.required` allows me to enable the add-on from about:addons, so we should be able to just append to `<ProfD>/prefs.js`: `user_pref("xpinstall.signatures.required", false);`

bin/run-utils.js
```diff
@@ -0,0 +1,192 @@
...
+function isErrorString(line) {
+  return /^\*{25}/.test(line) || /^\s*Message: [\D]*Error/.test(line);
+}
+
+var GARBAGE = [
+  /\[JavaScript Warning: "TypeError: "[\w\d]+" is read-only"\]/,
+  /JavaScript strict warning: /,
+  /\#\#\#\!\!\! \[Child\]\[DispatchAsyncMessage\]/,
+  /JavaScript strict warning: resource:\/\/\/modules\/sessionstore\/SessionStore\.jsm/,
+  /JavaScript strict warning: resource:\/\/gre\/components\/nsSearchService\.js/
```
Mardak: Doesn't the pattern on line 87 cover the ones on 89 and 90, so they're not necessary?

bin/run-utils.js
```diff
@@ -0,0 +1,192 @@
...
+function runFirefox(options) {
+  return When.promise(function(resolve, reject) {
+    options = options || {};
+
+    FxRunner({
+      "binary": options.binary,
+      "foreground": true,
+      "profile": options.profile,
+      env: Extend({}, process.env, FX_ENV),
+      verbose: true,
```
Mardak: This should probably be options.verbose?

bin/runfx.js
```diff
@@ -0,0 +1,88 @@
...
+var When = require("when");
+
+var DEFAULT_PROFILE = "loop-dev";
+
+Commander
+  .version(Version)
+  .option("-b, --binary <path>", "Path of Firefox binary to use.")
+  .option("-p, --profile <path>", "Path or name of Firefox profile to use.")
+  .option("--binary-args <CMDARGS>", "Pass additional arguments into Firefox.")
+  .parse(process.argv);
```
Mardak, line 27: Do we need to have an option for verbose here?

bin/run-utils.js
mikedeboer: Indeed :) Copy-paste error.

bin/runfx.js
mikedeboer, line 27: I can't think of a real reason why you'd want to suppress messages from the process landing into your console, so I left it on.

bin/runfx.js
mikedeboer: Ugh, that's possible? Ok, will add a check. Or a mkdir -p.

bin/run-utils.js
```diff
@@ -0,0 +1,192 @@
...
+      });
+
+      firefox.on("close", resolve);
+
+      firefox.stderr.setEncoding("utf8");
+      firefox.stderr.on("data", function(data) {
+        // Only print out annoying warnings if verbose is on
+        if (/^\s*System JS : WARNING/.test(data) && options.verbose) {
+          writeLog(data, "warn");
+        } else if (options.verbose) {
```
Mardak: It looks like these `options.verbose` checks always end up being false but we /do/ have FxRunner verbose: true always.

mikedeboer: True! Thanks for spotting this.

mikedeboer:
Did you point the command to run in Fx 45 or higher? You can specify this by specifying `-b Nightly` or to a `-b /path/to/bin/firefox` binary directly.

mikedeboer:
Ah, I'm also seeing that you need to enable the add-on on Nightly explicitly in the add-on manager.
Can we move ahead with this minor nuisance? I'm a bit wary about setting that pref in a profile...

bin/runfx.js
```diff
@@ -0,0 +1,100 @@
...
+        });
+      } else {
+        // Removing old symlink or file.
+        Fs.unlink(path).then(resolve).catch(reject);
+      }
+    }).catch(resolve);
+  });
+}
+
+function onExit() {
```
mikedeboer, line 49: Perhaps I should rename this to `exitWithError`?

Mardak:
@Standard8 should we initially land runfx in a state that requires some one-time pref toggling and add-on enabling? Common case subsequent runs should work fine with the current pull request.

mikedeboer:
FWIW, I never had to flip a pref.

Standard8:
As mentioned on irc I think its fine for this to land without pref setting and we can do that in a follow-up. Just having the option to start FF with the add-on is already a great improvement.

bin/runfx.js
```diff
@@ -0,0 +1,100 @@
...
+          // Add-on should be in the correct place, so now we can run Firefox.
+          return Util.runFirefox(Commander).then(function() {
+            throw "Firefox terminated properly";
+          }).catch(function() {
+            console.log("Removing symlink");
+            Fs.unlink(addonTargetDir);
+          });
+        });
+      }).catch(onExit);
+    }).catch(onExit);
```
Mardak: There's a lot of nesting and hard-to-match-catches that seems unnecessary given the use of Promises. I made a commit that should be pretty much equivalent: https://github.com/Mardak/loop/commit/1de36d268f3fed283638c2be11572b7a5e2044e8

mikedeboer:
@Mardak I like your changes!

mozilla-autolander-deprecated:

Mardak:
Squashed commits and merged as https://github.com/mozilla/loop/commit/c19b20deedae7d724705eaa13121190cd6c50e84

Mardak:
@mikedeboer fyi, I updated the author of the commit to match your other commits. Not sure if you accidentally/intentionally had the commits with info@

PR #10, opened 2015-12-14
Bug 1232441 - Add panel tests to karma, get the shared tests running in Chrome when run locally.

Standard8:


package.json
```diff
@@ -24,11 +24,12 @@
     "expose-loader": "0.7.x",
     "express": "4.13.3",
     "imports-loader": "0.6.5",
-    "nodemon": "1.8.1",
     "karma": "^0.12.37",
+    "karma-chrome-launcher": "^0.2.2",
```
dmose: Please add an exact version (e.g. no preceding ^) so that we keep things more deterministic.  Feel free to also remove the other ^ lines in package.json while you're at it.  :-)

dmose:
If you don't want to do it here, can you file a followon bug to fix the warning spew from some of the tests inside karma?

.travis.yml
```diff
@@ -20,6 +22,11 @@ install:
   - npm install
 
 before_install:
+  - wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+  - sudo sh -c 'echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
```
dmose, line 28: This looks like it could be fragile if the vm format shifts much, but maybe that's not an issue.  A comment explaining what the above line (or maybe the whole thing) is doing would be nice.

.travis.yml
```diff
@@ -20,6 +22,11 @@ install:
...
+  - sudo apt-get update -qq
```
dmose, line 30: It's not particularly clear why this makes sense.  I'm guessing it's because that command happens to look in the file you just appended to?  Is there a way to do this with a command-line switch rather than using an append side-effect in order to make it more readable?  If not, commentary could help here too.

shared/test/loop_mocha_utils.js
```diff
@@ -83,6 +83,12 @@ var LoopMochaUtils = (function(global, _) {
     });
   };
 
+  syncThenable.resolve = function(result) {
```
dmose: Why doesn't this get overwritten by the "this.resolve" dynamic function assignment inside the constructor?

As a optional nit, it would be nifty to rename syncThenable to SyncThenable, since it's a constructor.

shared/test/loop_mocha_utils.js
```diff
@@ -201,7 +207,7 @@ var LoopMochaUtils = (function(global, _) {
    *                             to the listeners.
    */
   function publish() {
-    var args = Array.slice(arguments);
+    var args = Array.prototype.slice.call(arguments);
```
dmose, line 226: A brief comment about why this is necessary would be nice.

test/karma/head.js
```diff
@@ -1,11 +1,8 @@
...
-// Used for desktop coverage tests because triggering methods on
-// DOMContentLoaded proved to lead to race conditions.
-
-sinon.stub(document, "addEventListener");
-console.log("[head.js] addEventListener stubbed to prevent race conditions");
-
 document.body.appendChild(document.createElement("div")).id = "fixtures";
 console.log("[head.js] div#fixtures added to attach DOM elements");
+
+// Trap any errors and warnings that are raised.
```
dmose: Maybe also "in preparation for a later call to addErrorCheckingTests"

shared/test/loop_mocha_utils.js
Standard8, line 226: This is the "proper" way to call it:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Array-like_objects

The other one works in FF, but I don't think its as official as this one.

shared/test/views_test.js
```diff
@@ -758,7 +753,18 @@ describe("loop.shared.views", function() {
 
       expect(element).not.eql(null);
       expect(element.className).eql("local-video");
-      expect(element.muted).eql(true);
+
+      // Google Chrome doesn't seem to set "muted" on the element at creation
+      // time, so we need to do the test as async.
```
dmose, line 742: Weird.  I wonder if this is really React papering over some difference in Chrome.

test/karma/tail.js
```diff
@@ -0,0 +1,15 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Stop the default init functions running to avoid conflicts in tests
```
dmose: Isn't this really an indicator that we want to be using dependency injection somewhere that we're not?  I don't think it's necessarily required to fix that before landing this, but if you agree, it might be worth adding an XXX comment.

shared/test/loop_mocha_utils.js
Standard8: > Why doesn't this get overwritten by the "this.resolve" dynamic function assignment inside the constructor?

The resolve function doesn't get created until the SyncThenable is created. So, in some places the code is doing `Promise.resolve()` (note: not constructing the promise), this equates to `SyncThenable.resolve()` which doesn't currently exist. Hence, we need to create it.

test/karma/tail.js
Standard8: The listeners are injected in the global scope as we can't call them direct (something's got to kick everything off). Therefore using dependency injection is difficult. I think this method is slightly better than stubbing addEventListener globally.

Standard8:
> If you don't want to do it here, can you file a followon bug to fix the warning spew from some of the tests inside karma?

I fixed some of these and filed [bug 1233739](https://bugzilla.mozilla.org/show_bug.cgi?id=1233739) on the rest.

shared/test/loop_mocha_utils.js
dmose, line 226: I know why it's there, I meant a brief comment in the code, since non-expert JS coders might well assume that arguments is an Array.  :-)

test/karma/tail.js
dmose: Add this as a comment to the code?

shared/test/loop_mocha_utils.js
```diff
@@ -85,15 +85,21 @@ var LoopMochaUtils = (function(global, _) {
     });
   };
 
+  SyncThenable.resolve = function(result) {
+    return new SyncThenable(function(resolve) {
+      resolve(result);
+    });
+  };
+
```
dmose, line 106: Please add a comment to each "resolve"  function (both the dynamic and static versions in this file) explaining when each one gets used for clarity.

dmose:
Looks good, r=dmose with comment changes addressed.  Thanks!

PR #11, opened 2015-12-15
Bug 1232030 - Rebuild npm install if package.json changed

chrafuse:


Makefile
```diff
@@ -19,11 +19,12 @@ NODE_LOCAL_BIN := ./node_modules/.bin
...
-install: npm_install
-	pip install -r virtualenv
+.PHONY: install
+install: node_modules
+	pip install -r require.pip
 
-npm_install:
-	@npm install
+node_modules: package.json
+	npm install
```
Standard8, line 27: Nice. One minor change I think we need is to add a line to `@touch node_modules` after the npm install.

The reason is (and I just had this locally), that you can have a node_modules already installed, and you change branch so package.json gets "touched" but node_modules is already up to date... so then make thinks it needs to rebuild it again, but `npm install` doesn't change anything and tries to do it each time.

PR #12, opened 2015-12-15
Bug 1232111 - Update the README file with a new skeleton and add a contributing doc.

Standard8:

PR #13, opened 2015-12-15
Make sure to make ./node_modules/ so packages are installed locally

ianb:
r? @chrafuse

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ install: node_modules
 	pip install -r require.pip
 
 node_modules: package.json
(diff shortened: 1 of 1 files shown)


Mardak:
@Standard8 If we're merging via github interface, we should probably have r= on the merge commit?

mikedeboer:
Absolutely!

Standard8:
Yes, though I think we can let it go in this instance.

I've also made this a bit more explicit in the docs in PR #12.

PR #14, opened 2015-12-16
Bug 1231808 - Hide pause from Infobar

daveccrit:
Added type of pause and stop to buttons. Used selector in CSS to target button type of pause and hide

PR #15, opened 2015-12-16
Revert "Bug 1231808 - Hide pause from Infobar"

mikedeboer:
Reverts mozilla/loop#14

@Standard8


Standard8:
r=Standard8

mikedeboer:
Now, Mark, this PR was created with GH's 'revert' button, so no r=... will be shown. What's our policy going to be here?

PR #18, opened 2015-12-16
Bug 1230058 - remove unused string properties

chrafuse:


dmose:
Looks good to me, r=dmose.  Please amend the commit to include that before merging.

mozilla-autolander-deprecated:

PR #22, opened 2015-12-18
Bug 1232793 - Automatically update npm modules using Updtr

chrafuse:
Updtr output and modification after testing each module for latest version compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,16 @@
     "node": "0.10.39",
     "npm": "2.14.1"
   },
-  "dependencies": {},
+  "dependencies": {
+    "babel": "^5.8.34",
+    "classnames": "^2.2.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -13,7 +13,16 @@
     "node": "0.10.39",
     "npm": "2.14.1"
   },
-  "dependencies": {},
+  "dependencies": {
+    "babel": "^5.8.34",
+    "classnames": "^2.2.1",
```
Standard8, line 18: We need to use fixed versions rather than caret ranges, to ensure builds can be reproduced exactly and easily.

Additionally, most of the things you've listed here are dev dependencies rather than what we're currently calling "production". They also duplicate things in the dev dependencies section...

chrafuse:
PR superseded by #31

PR #25, opened 2015-12-18
Bug 1231808 - Hide Pause button in Notification bar

daveccrit:
Bug 1231808 - Hide pause from Infobar


mozilla-autolander-deprecated:

PR #28, opened 2015-12-21
Bug 1232684 - Use makeFile to generate chrome.manifest from jar.mn

chrafuse:


Makefile
```diff
@@ -145,6 +145,8 @@ standalone: node_modules
 .PHONY: add-on
 add-on: node_modules
 	mkdir -p $(BUILT)/$@
+	sed -n -e '/^%/p' $@/jar.mn | sed -e "s/^% //g" | sed -e "s/%/chrome\//g" > $@/chrome.manifest
```
Standard8: Can you separate this out to a separate rule please? No point in regenerating it each time - see some of the vendor library handling for instance.

Also, this should output direct to `$(BUILT)/add-on/chrome.manifest`, and we should git rm the existing`add-on/chrome.manifest`.

Makefile
```diff
@@ -145,6 +145,8 @@ standalone: node_modules
...
+	sed -e '/^%/d' $@/jar.mn > $(BUILT)/$@/jar.mn
```
Standard8: We don't need to generate the jar.mn here, that's just the source (we need it for when we export back to m-c).

Makefile
```diff
@@ -145,7 +146,7 @@ standalone: node_modules
 .PHONY: add-on
 add-on: node_modules
 	mkdir -p $(BUILT)/$@
-	$(RSYNC) $@/chrome.manifest $@/chrome/bootstrap.js $(BUILT)/$@
+	$(RSYNC) $@/jar.mn $@/chrome/bootstrap.js $(BUILT)/$@
```
Standard8: We don't need the jar.mn here - just remove it, then add `$(BUILT)/$(ADD-ON)/chrome.manifest` as a dependency on the `add-on:` line above.

Makefile
```diff
@@ -164,6 +165,13 @@ add-on: node_modules
 	mkdir -p $(BUILT)/$@/chrome/locale/en-US
 	cat locale/en-US/$@.properties locale/en-US/shared.properties > $(BUILT)/$@/chrome/locale/en-US/loop.properties
 
+.PHONY: chrome-manifest
+chrome-manifest: $(ADD-ON)/jar.mn
```
Standard8: No need for the .PHONY as this is a real dependency. We do however need to change this line to:

`$(BUILT)/$(ADD-ON)/chrome.manifest: $(ADD-ON)/jar,mn`

that sets up the output chrome.manifest as dependent on jar.mn.

add-on/chrome.manifest
```diff
@@ -1,24 +0,0 @@
-content loop chrome/content/ contentaccessible=yes
-locale loop en-US chrome/locale/en-US/
```
Standard8, line 2: Please can you add a line to jar.mn to replace this. I'd forgotten about it before. You should just be able to add:

`% locale loop en-US %locale/en-US/`

If you could put that under the content line in add-on/jar.mn that'd be great

Standard8:
Looks good, r=Standard8 with the additional comment fixed.

chrafuse:
Added locale line from comments and pushed to mozilla master.

PR #34, opened 2015-12-29
Bug 1070208 - pass fxAProfileUid as query param to fxa settings page so it open the right account

Vidhuran:


mozilla-autolander-deprecated:

add-on/chrome/test/mochitest/browser_fxa_login.js
```diff
@@ -456,8 +470,10 @@ add_task(function* openFxASettings() {
         }
         gBrowser.removeTabsProgressListener(progressListener);
         let contentURI = Services.io.newURI(params.content_uri, null, null);
-        is(aBrowser.currentURI.spec, Services.io.newURI("/settings", null, contentURI).spec,
+        is(aBrowser.currentURI.spec, Services.io.newURI("/settings?uid=abcd1234", null, contentURI).spec,
            "Check settings tab URL");
+        Services.prefs.clearUserPref("loop.fxa_oauth.profile");
+        Services.prefs.clearUserPref("loop.fxa_oauth.tokendata");
```
Standard8: Please can you make the clearUserPref calls be in a registerCleanupFunction handler. There's an example here: 

https://github.com/mozilla/loop/blob/918a7d1a22aa8274239f2cf448c236b78759bdff/add-on/chrome/test/mochitest/browser_fxa_login.js#L54

We want registerCleanupFunction called within this test (openFxASettings) - just after where they are set would be a good place.

Vidhuran:
Updated PR as per the review comments

Vidhuran:
Sorry, I've now removed the redundant lines

PR #35, opened 2015-12-29
Bug 1234905 - Upgrade React to 0.14.8

chrafuse:
Added dependencies.  Replacement of new methods are to be done still.


add-on/panels/js/conversation.jsx
```diff
@@ -195,7 +195,7 @@ loop.conversation = (function(mozL10n) {
         textChatStore: textChatStore
       });
 
-      React.render(
+      ReactDOM.render(
```
Standard8, line 220: You need to change the stub in conversation_test.js to account for this change. There's probably similar stubs that may need updating in panels_test.js and webapp_test.js (and maybe other places)

shared/test/views_test.js
```diff
@@ -604,7 +604,7 @@ describe("loop.shared.views", function() {
       });
 
       it("should not throw if the element is not a video object", function() {
-        sinon.stub(view, "getDOMNode").returns({
+        sinon.stub(ReactDOM.findDOMNode(view)).returns({
```
Standard8: This change doesn't work as you're not actually stubbing anything - you're just giving sinon.stub an object, and not a second argument.

You need to change this (and the other 3 instances) to:

```
sinon.stub(ReactDOM, "findDOMNode").returns(...
```

Standard8: Oh, and these need to become sandbox.stub since we're stubbing a "global" rather than a created item.

shared/test/linkifiedTextView_test.js
```diff
@@ -37,7 +37,7 @@ describe("loop.shared.views.LinkifiedTextView", function() {
 
   describe("LinkifiedTextView", function() {
     function renderToMarkup(string, extraProps) {
-      return React.renderToStaticMarkup(
+      return ReactDOM.render(
```
Standard8: This isn't an equivalent change. We were rending to static markup previously, and now we're rending as an object.

For these tests, I think its best if we do test against the static markup - testing the dom tree would be complicated.

Therefore I think just for this shared test, we could include react-dom-server to do that rendering.

add-on/panels/js/panel.jsx
```diff
@@ -128,16 +128,16 @@ loop.panel = (function(_, mozL10n) {
       var locale = mozL10n.getLanguage();
       var tosHTML = mozL10n.get("legal_text_and_links3", {
         "clientShortname": mozL10n.get("clientShortname2"),
-        "terms_of_use": React.renderToStaticMarkup(
+        "terms_of_use": "<a>Fake</a>", /* ReactDOM.render(
```
Standard8: I've had a chat to the L10n folks, and it seems like our best option may be to upgrade our l10n-*.js as then we'd able to use constructs such as:

```
legal_text="You agree to the <a>terms_of_use</a>"
```

in the L10n file, and be able to specify attributes for the element. I'm not 100% sure it'll work properly for two `<a>` replacements in the same string, but we'll have to take that as it comes.

The bug for this is https://bugzilla.mozilla.org/show_bug.cgi?id=1230208 - and I think we'll have to block this bug on at least investigating that. It feels like the upgrade of L10n.js is going to need some testing and investigation so we're going to need to schedule that in.

Standard8: Scrub the previous comment about upgrading L10n.js. 

There's a simpler way. This section change be changed to:

```
  var tosString =
    '<a href="' + this.state.terms_of_use_url + '" target="_blank">' +
    mozL10n.get("legal_text_tos") +
    "</a>";

  var privacyString =
    '<a href="' + this.state.privacy_notice_url + '" target="_blank">' +
    mozL10n.get("legal_text_privacy") +
    "</a>";

  var tosHTML = mozL10n.get("legal_text_and_links3", {
    "clientShortname": mozL10n.get("clientShortname2"),
    "terms_of_use": tosString,
    "privacy_notice": privacyString
  });
```

It functions the same, just doesn't take advantage of the static markup rending that react has. However, I think that's reasonable to loose considering the react changes.

There's also a similar ToS link setup in the standalone that can be changed in the same manner.

PR #36, opened 2015-12-31
Bug 1226607 - When there's no converstions - the UX isn't matching NUJ

chrafuse:


add-on/panels/js/panel.jsx
```diff
@@ -755,10 +737,6 @@ loop.panel = (function(_, mozL10n) {
         return this._renderLoadingRoomsView();
       }
 
-      if (!this.state.rooms.length) {
-        return this._renderNoRoomsView();
```
Mardak: This logic was short circuiting the below. Do you have a screenshot of what no rooms looks like now? I would guess it would show the Recently Browsed header text with this removal.

add-on/panels/js/panel.jsx
```diff
@@ -718,22 +718,14 @@ loop.panel = (function(_, mozL10n) {
...
-                {mozL10n.get("no_conversations_message_heading2")}
-              </p>
-              <p className="panel-text-medium">
-                {mozL10n.get("no_conversations_start_message2")}
-              </p>
-            </div>
-          </div>
-        </div>
-      );
+    _renderRoomsTitle: function() {
```
Mardak: Probably rename this to _renderRoomsListTitle as RoomsTitle almost sounds like the titles for the rooms.

add-on/panels/js/panel.jsx
```diff
@@ -718,22 +718,14 @@ loop.panel = (function(_, mozL10n) {
...
+      if (this.state.rooms.length) {
```
Mardak: We can flip this check and return null early.

add-on/panels/js/panel.jsx
```diff
@@ -745,6 +737,29 @@ loop.panel = (function(_, mozL10n) {
       );
     },
 
+    _renderRoomList: function() {
```
Mardak: _renderRoomsList (add s)

add-on/panels/js/panel.jsx
```diff
@@ -745,6 +737,29 @@ loop.panel = (function(_, mozL10n) {
...
+      if (this.state.rooms.length) {
```
Mardak: Again flip check and explicit `return null`

add-on/panels/js/panel.jsx
```diff
@@ -745,6 +737,29 @@ loop.panel = (function(_, mozL10n) {
...
+                }
+
+              return (
+              <RoomEntry
+                dispatcher={this.props.dispatcher}
+                isOpenedRoom={room.roomToken === this.state.openedRoom}
+                key={room.roomToken}
+                room={room}/>
+                );
+              }, this)
```
Mardak: You have some odd indenting going on several places above that'll also need to be adjusted when flipping the condition.

add-on/panels/js/panel.jsx
```diff
@@ -755,32 +770,11 @@ loop.panel = (function(_, mozL10n) {
...
-                <RoomEntry
-                  dispatcher={this.props.dispatcher}
-                  isOpenedRoom={room.roomToken === this.state.openedRoom}
-                  key={room.roomToken}
-                  room={room} />
-              );
-            }, this)
-          }</div>
+          {this._renderRoomsTitle()}
+          {this._renderRoomList()}
```
Mardak: Actually, why split this into 2 separate calls instead of just one _renderRoomsList that handles the title and list?

add-on/panels/test/panel_test.js
```diff
@@ -944,22 +944,6 @@ describe("loop.panel", function() {
       sinon.assert.calledOnce(fakeWindow.close);
     });
 
-    it("should render the no rooms view when no rooms available", function() {
-      var view = createTestComponent();
-      var node = view.getDOMNode();
-
-      expect(node.querySelectorAll(".room-list-empty").length).to.eql(1);
-    });
```
Mardak: We should still check for what should be handled when there's no rooms available.

add-on/panels/js/panel.jsx
chrafuse: Originally the mark-up was in the render and would break as soon as I added conditionals and moved elements.  Created functions as they are allowed in the render and not so much the local variables (var localVar) in the render.

add-on/panels/js/panel.jsx
```diff
@@ -777,10 +752,29 @@ loop.panel = (function(_, mozL10n) {
                   dispatcher={this.props.dispatcher}
                   isOpenedRoom={room.roomToken === this.state.openedRoom}
                   key={room.roomToken}
-                  room={room} />
+                  room={room}/>
```
Mardak: nit: keep the space before />

add-on/panels/js/panel.jsx
```diff
@@ -718,22 +718,14 @@ loop.panel = (function(_, mozL10n) {
...
-      return (
-        <div className="rooms">
-          {this._renderNewRoomButton()}
-          <div className="room-list-empty">
-            <div className="no-conversations-message">
-              <p className="panel-text-medium">
-                {mozL10n.get("no_conversations_message_heading2")}
-              </p>
-              <p className="panel-text-medium">
-                {mozL10n.get("no_conversations_start_message2")}
```
Mardak: These strings need to be removed from add-on.properties

add-on/panels/js/panel.jsx
```diff
@@ -718,22 +718,14 @@ loop.panel = (function(_, mozL10n) {
...
-              </p>
-            </div>
-          </div>
-        </div>
-      );
+    _renderRoomsListTitle: function() {
```
Mardak: nit: this method is alphabetically after _renderRoomsList, so put it there just before render.

add-on/panels/js/panel.jsx
```diff
@@ -777,10 +752,29 @@ loop.panel = (function(_, mozL10n) {
...
+      }
+
+      if (this.state.pendingInitialRetrieval) {
+        return this._renderLoadingRoomsView();
+      }
+
+      return (
+        <div className="rooms">
+          {this._renderNewRoomButton()}
+          {this._renderRoomsListTitle()}
```
Mardak: I still don't see why these need to be in separate methods. You can do `{this.state.rooms.length ? <h1>{mozL10n..}</h1> : null}`. You could stash the rooms.length in some variable, e.g., haveRooms and do `{haveRooms ? ... : null}` for both the title and rooms list.

See https://github.com/mozilla/loop/blob/master/add-on/panels/js/panel.jsx#L490

PR #37, opened 2015-12-31
Bug 1232663 - Call closeWindow method after sign in/out to prevent panel from staying open

Vidhuran:


add-on/panels/test/panel_test.js
```diff
@@ -212,7 +212,7 @@ describe("loop.panel", function() {
       var view = TestUtils.renderIntoDocument(
         React.createElement(loop.panel.SettingsDropdown));
 
-      expect(view.getDOMNode().querySelectorAll(".icon-account"))
+      expect(view.getDOMNode().querySelectorAll(".entry-settings-account"))
```
Standard8, line 215: Nice catch!

add-on/panels/test/panel_test.js
```diff
@@ -350,50 +350,94 @@ describe("loop.panel", function() {
...
-          GetUserProfile: function() { return { email: "test@example.com" }; }
+      describe("UserLoggedIn", function() {
+        it("should show a signout entry when user is authenticated", function() {
+          loop.storedRequests.GetUserProfile = { email: "test@example.com" };
+
+          mountTestComponent();
+
+          sinon.assert.calledWithExactly(document.mozL10n.get,
+              "settings_menu_item_signout");
+          sinon.assert.neverCalledWith(document.mozL10n.get,
```
Standard8: This isn't your fault as its existing code, but can we improve these tests whilst we're touching them? Using a test for a function call when testing if a button is displayed or not isn't ideal.

Can you drop the two sinon.assert calls and replace them with:

```
var view = mountTestComponent();

expect(view.getDOMNode().querySelectorAll(".entry-settings-signout"))
  .to.have.length.of(1);
expect(view.getDOMNode().querySelectorAll(".entry-settings-signin"))
  .to.have.length.of(0);
```

add-on/panels/test/panel_test.js
```diff
@@ -350,50 +350,94 @@ describe("loop.panel", function() {
...
 
-        sinon.assert.calledWithExactly(document.mozL10n.get,
-                                       "settings_menu_item_settings");
-      });
+          mountTestComponent();
 
-      it("should open the FxA settings when the account entry is clicked",
-         function() {
-           loop.storedRequests.GetUserProfile = { email: "test@example.com" };
+          sinon.assert.calledWithExactly(document.mozL10n.get,
```
Standard8: This wants to be replaced by something similar as well (`.entry-settings-account`)

Vidhuran:
Sure, I'll improve those tests as you say.

add-on/panels/test/panel_test.js
```diff
@@ -952,12 +982,10 @@ describe("loop.panel", function() {
...
     it("should call mozL10n.get for room empty strings", function() {
-      createTestComponent();
+      var view = createTestComponent();
 
-      sinon.assert.calledWithExactly(document.mozL10n.get,
-                                     "no_conversations_message_heading2");
-      sinon.assert.calledWithExactly(document.mozL10n.get,
-                                     "no_conversations_start_message2");
+      expect(view.getDOMNode().querySelectorAll(".room-list-empty").length).to.eql(1);
+      expect(view.getDOMNode().querySelectorAll(".panel-text-medium").length).to.eql(2);
```
Vidhuran, line 988: This will change soon as there are some bugs that want this behavior to be changed.

PR #42, opened 2016-01-05
Include some information in the README about building the add-on and using runfx

ianb:


mozilla-autolander-deprecated:

README.md
```diff
@@ -8,6 +8,31 @@ window.
...
+```shell
+$ ./bin/runfx.js -p nightly -p ./loop-dev/
+```
+
+**One time only**: This will use a new profile created in a `./loop-dev/`
+directory.  When Firefox opens you must go to the URL `about:config` and
+look for the setting `xpinstall.signatures.required`.  This must be set to
+**false**.  Then go to
+**Tools &gt; Add-ons** and find the Firefox Hello addon.  You must manually
+re-enable the add-on.
```
Standard8, line 34: nit: I think its reasonable just to say "enable".

Standard8:
I think ideally I'd like this to be split out to a separate file, as we're going to be needing to add a lot more to it - for building, running, standalone, testing etc. However, if you want to land it as-is for now, and we'll move it later, I'm fine with that.

dmose:
Shouldn't that be '-b nightly -p ...'?

mozilla-autolander-deprecated:

ianb:
I'll infer @Standard8's response as an r+ (with the -b fix)

PR #45, opened 2016-01-05
Fix the profile creation documentation

ianb:
In previous pull request I offered no good way to create the profile, this fixes that.


mozilla-autolander-deprecated:

Standard8:
The way I've always used on each platform is:

`/path/to/firefox -createProfile loop-dev`

On Mac this becomes something like:

`/Applications/FirefoxNightly.app/Contents/MacOS/firefox -createProfile loop-dev`

dmose:
That will create a profile in the normal profile location, but I haven't yet been able to cause it to create a profile in the current directory.

Path of least resistance may be to just recommend people use a profile in the normal place.

ianb:
Both work – if you create a named profile then use it with -p name that works, or a directory.  I thought that Firefox looked for `/` to see if it's a directory name.  So when I do `-Profile ./loop-dev/` then I get a directory (but I haven't tried/wasn't aware of -createProfile)

dmose:
Interestingly,

``` shell
$ open /Applications/FirefoxNightly.app --args -Profile ./loop-dev/
```

As well as a bunch of different variants, always generates https://monosnap.com/file/GeqZj1o3MAt586OcMwOFhMimTGR99n# for me.

I'm beginning to wonder if my profile registry is corrupt or something

ianb:
I had some trouble with open/--args once, though then it worked again later and I couldn't figure out what I changed so I thought I had imagined it.  Now it is causing problems again for me (same as for @dmose), while `/Applications/FirefoxNightly.app/Contents/MacOS/firefox --args -Profile ./test-profile/` works fine.

ianb:
Stuff is hard.  Here's how `-createProfile` works: https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#-CreateProfile_.22profile_name_profile_dir.22

The profile created is super-minimal.  I had luck putting this line in `profiledir/prefs.js`:

```
user_pref("xpinstall.signatures.required", false);
```

The add-on was still disabled when I started Firefox with runfx, but I could manually enable it.

So maybe a script that would do everything would be:

``` shell
/Applications/FirefoxNightly.app/Contents/MacOS/firefox -createProfile "loop-test-profile $(pwd)/test-profile/"
echo 'user_pref("xpinstall.signatures.required", false);' >> test-profile/prefs.js
./bin/runfx.js -b nightly -p ./test-profile/
```

README.md
```diff
@@ -11,10 +11,17 @@ web server.
...
-you can run:
+**One time only**: We recommend you create a new profile for Firefox.  You can do this by finding the Firefox executable and running it with the `-ProfileManager` option, or `-Profile ./loop-dev/` (to create a profile in `./loop-dev/`).  On a Mac you can do this like:
+
+```shell
+$ open /Applications/FirefoxNightly.app --args -Profile ./loop-dev/
+```
+
+When Firefox opens you must go to the URL `about:config` and
+look for the setting `xpinstall.signatures.required`.  This must be set to
+**false**.  If you forget to do this then later the Hello icon will not appear and the add-on may be disabled; if this happens to go **Tools &gt; Add-ons** and find the Firefox Hello add-on, and enable it.
```
dmose, line 22: This wants to be Tools > Add-ons > Extensions, I think.

README.md
```diff
@@ -11,10 +11,17 @@ web server.
...
-all the normal Unix command-line utilities, and have installed Node and npm,
-you can run:
+**One time only**: We recommend you create a new profile for Firefox.  You can do this by finding the Firefox executable and running it with the `-ProfileManager` option, or `-Profile ./loop-dev/` (to create a profile in `./loop-dev/`).  On a Mac you can do this like:
+
+```shell
+$ open /Applications/FirefoxNightly.app --args -Profile ./loop-dev/
+```
+
+When Firefox opens you must go to the URL `about:config` and
+look for the setting `xpinstall.signatures.required`.  This must be set to
```
dmose, line 21: As of today, you also have to set xpinstall.signatures.required to false, and even that still leaves the thing disabled by default.

dmose, line 21: Er, I meant, xpinstall.whitelist.required to false.

dmose:
jpm itself actually depends on this npm library for profile creation:

https://github.com/saadtazi/firefox-profile-js

I didn't catch all the details from @mikedeboer in this AMs standup, but I wonder if we don't want to just use that.

Standard8:
@dmose We already do rely on that package - see package.json/run-utils.js

README.md
Standard8, line 21: I've never needed to set this, reading the code, I don't think you would need to set it: http://hg.mozilla.org/mozilla-central/annotate/8a7cc8b7aa55/browser/components/preferences/in-content/security.js#l38

ianb:
I'm pretty sure the process has changed since the PR was made, and now there is a section about how it works now.

PR #46, opened 2016-01-05
Some Makefile improvements:

ianb:
- Add `make help`
- Mark several rules as PHONY
- Use cp instead of rsync in some simple/obvious cases (single files)
- Call directly into virtualenvs instead of using activate
- Use python -m to load mozbuild.preprocessor
- Remove some commands from the add-on target, to make them dependencies with rules (so the commands are done only when needed)


mozilla-autolander-deprecated:

dmose:
It's not obvious to me why switching from rsync to cp is worthwhile, rather than just sticking with one consistent tool.

ianb:
I'd like to see rsync disappear entirely – make with the appropriate rules _is_ rsync.  Those single-file cases were the easiest to understand and remove.  Maybe cp is also lighter than rsync, but I don't really know.

In the case of cp specifically the mkdirs are really noisy and boring, since you run it for every single file.  But blocking all seems fine too.

mozilla-autolander-deprecated:

dmose:
The reason I chose rsync (which I should have commented) is that we've had issues cause by the sometimes confusing syntax inconsistencies between BSD (Mac) and Linux cp argument semantics.

r+ if you revert the cp changes (and feel free to copy the above info in as a comment above the RSYNC definition)

ianb:
I don't think we need any cp options, do we?  Otherwise I would assume that cp without options would work the same on Linux or Mac.  My concern with rsync is that it lets you do something (recursive copying) which make can't understand.

dmose:
It's not the options, it's how it handles trailing arguments differently depending on what sort of files/directories are already in place.  rsync reliably does the same thing and thus is less of a maintenance headache.  I don't recall the details; @Standard8 might.

I don't really understand your concern about rsync.  Can you elaborate in more detail about what negative fallout this might have?

Standard8:
I've a feeling they are both sometimes just as sensitive. For me, rsync seems slightly nicer as you don't have to be explicit about `mkdir` then `cp`. We should probably have some arguments to cp, like `-r` and `-p`.

I guess I don't have too stronger feeling here.

ianb:
I'm just assuming this has bitrotted too much to be worth doing any more

PR #48, opened 2016-01-06
Bug 1236082 - Increase user ID email display in panel footer

chrafuse:


add-on/panels/js/panel.jsx
```diff
@@ -340,7 +340,7 @@ loop.panel = (function(_, mozL10n) {
       if (this.props.userProfile && this.props.userProfile.email) {
         return (
           <div className="user-identity">
-            {loop.shared.utils.truncate(this.props.userProfile.email, 24)}
+            {loop.shared.utils.truncate(this.props.userProfile.email, 40)}
```
Standard8: I just happened to flick on to this PR... This landed in [bug 1183636](https://bugzilla.mozilla.org/show_bug.cgi?id=1183636), but I don't understand why we've got a special truncate function rather than using layout with `text-overflow: ellipsis;`.

@mikedeboer do you remember the reasoning behind this?

mikedeboer: I can't remember if there was one. I _think_ it was due to some kind of restraint with the layout (since the value is so specific).
Anyway, I think it's better to try using CSS to fix this.

add-on/panels/css/panel.css
```diff
@@ -655,6 +655,9 @@ html[dir="rtl"] .settings-menu .dropdown-menu {
   color: #000;
   font-weight: bold;
   margin: 0;
+  width: 270px;
```
Mardak: Did this 270px come from the design spec? If not, a margin/padding end might be better than a specific width as the footer width is set to 100%.

add-on/panels/js/panel.jsx
```diff
@@ -340,7 +340,7 @@ loop.panel = (function(_, mozL10n) {
       if (this.props.userProfile && this.props.userProfile.email) {
         return (
           <div className="user-identity">
-            {loop.shared.utils.truncate(this.props.userProfile.email, 24)}
+            {this.props.userProfile.email}
```
dmose, line 347: Does this also mean we can get rid of truncate?

dmose, line 347: As well as any relevant tests.

chrafuse, line 347: Got rid of the outer user-details containing DIV and the overflow works correctly because there is a single element to overflow now.

add-on/panels/css/panel.css
```diff
@@ -610,15 +600,32 @@ html[dir="rtl"] .settings-menu .dropdown-menu {
...
+  flex: 1;
   color: #000;
   font-weight: bold;
   margin: 0;
+  padding-right: 1rem;
+  overflow-x: hidden;
+  text-overflow: ellipsis;
+}
+
+html[dir="rtl"] .footer > .user-identity {
```
dmose: Instead of the duplication of rules, please use padding*end (either inline or block, I'm not sure which).

add-on/panels/js/panel.jsx
dmose, line 347: Please remove both truncate and the tests for it, since this is the only user.

add-on/panels/css/panel.css
```diff
@@ -610,15 +600,32 @@ html[dir="rtl"] .settings-menu .dropdown-menu {
...
+  padding-left: 1rem;
+}
+
+.footer > .user-identity > .dropdown-menu {
```
dmose: Please remove the .user-identity > .dropdown-menu stuff, as we no longer have a menu here for availability status.

dmose:
The new changes have a visual problem: the email address is no longer in the same vertical line as the settings bar.

dmose:
Or, rather as the settings gear.

chrafuse:
Used margin-inline-end instead of padding-block-end.

add-on/panels/css/panel.css
```diff
@@ -610,15 +600,19 @@ html[dir="rtl"] .settings-menu .dropdown-menu {
   height: 42px;
 }
 
-.footer .signin-details {
+.footer > .signin-details {
   align-items: center;
   display: flex;
```
dmose: Display flex can be removed, as there are no flexing children.

add-on/panels/css/panel.css
```diff
@@ -610,15 +600,19 @@ html[dir="rtl"] .settings-menu .dropdown-menu {
...
 }
 
-.footer .user-identity {
+.footer > .user-identity {
+  flex: 1;
```
dmose: Add an XXX comment "can we remove this?" here, or just remove it.

dmose:
r=dmose, once you've dropped display: flex

PR #57, opened 2016-01-07
Bug 1237781 - Fix unnecessary rebuild by simplifying Makefile

dmose:


Makefile
```diff
@@ -149,25 +149,19 @@ built_ui_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/ui/%.js)
...
+	mkdir -p $(@D)
 	$(BABEL) $< --out-file $@
 
-$(BUILT)/ui/shared/js:
-	@mkdir -p $@
-
 # We can't use $(shared_jsx_files) here because % rules don't accept that.
-$(BUILT)/ui/shared/js/%.js: shared/js/%.jsx $(BUILT)/ui/shared/js
+$(BUILT)/ui/shared/js/%.js: shared/js/%.jsx
+	@mkdir -p $(@D)
```
Mardak, line 164: This line has `@mkdir` while the rest are echo-y mkdir. I think it makes sense to suppress with @ as it could get spammy when building many jsx.

dmose, line 164: Whoops!  Just pushed up a commit fixing this.

2016-01-07 15:19 GMT-08:00 Ed Lee notifications@github.com:

PR #59, opened 2016-01-08
Bug 1236590 - create a loop-dev profile when no existing profile is specified and found

mikedeboer:
Create a new 'loop-dev' browser profile when none can be found or isn't specified. Also make sure that the right prefs have been set to aid in Hello developement

### DIFF
--- a/bin/run-utils.js
+++ b/bin/run-utils.js
@@ -127,7 +127,7 @@ function writeLog(data, type) {
  * @param {Object} options
  *   - `binary` path to Firefox binary to use
  *   - `profile` path to profile or profile name to use
- *   - `binaryArgs` binary arguments to pass into Firefox, split up by spaces
+ *   - `binaryArgs` binary arguments Array to pass into Firefox
  * @return {Promise}
  */
 function runFirefox(options) {
@@ -136,7 +136,7 @@ function runFirefox(options) {
 
     FxRunner({
       "binary": options.binary,
-      "foreground": true,
+      "foreground": ("foreground" in options) ? options.foreground : true,
       "profile": options.profile,
       env: Extend({}, process.env, FX_ENV),
       verbose: true,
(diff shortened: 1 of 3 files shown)


bin/run-utils.js
```diff
@@ -184,3 +185,215 @@ function runFirefox(options) {
...
+  "security.warn_submit_insecure": false,
+  "security.warn_viewing_mixed": false,
+  "security.warn_viewing_mixed.show_once": false,
+  "toolkit.telemetry.reportingpolicy.firstRun": false
+};
+
+function createEmptyProfile(profile) {
+  return When.promise(function(resolve, reject) {
+    var profileDir = randomString() + "." + profile;
+    var finder = new FxProfileFinder();
```
Mardak: Could we make use of the `-CreateProfile` command line argument instead of reimplementing some of the profile creation logic? Something like.. `require("child_process").spawn('/Applications/Firefox.app/Contents/MacOS/firefox-bin', ["-CreateProfile", "loop-dev"])`

mikedeboer: Sure, I didn't know about that one + didn't know that it shut down the process right after it created the profile. nice!

mikedeboer:
@Mardak I'm now using `-CreateProfile`, but it requires an update to the fx-runner module mozilla-jetpack/node-fx-runner#5

bin/run-utils.js
```diff
@@ -184,3 +184,195 @@ function runFirefox(options) {
...
+  if (value === "true") {
+    return true;
+  }
+  if (value === "false") {
+    return false;
+  }
+  return value;
+}
+
+function takePrefsSample(profilePath) {
```
Mardak, line 205: What do you mean by Sample? This looks to read in the whole prefs file as opposed to some of it. There's various things that make this function somewhat hard to read, and we might not actually need all the complexity anyway. See comment for setUserPrefs.

bin/run-utils.js
```diff
@@ -184,3 +184,195 @@ function runFirefox(options) {
...
+        }
+
+        return entry;
+      });
+
+      Object.getOwnPropertyNames(prefs).forEach(function(entry) {
+        if (seen[entry]) {
+          // An existing pref line has been overwritten above.
+          return;
+        }
```
Mardak, line 304: With all the above, it's nice to be clean and correctly replace existing prefs with the defaults, but `prefs.js` file works just fine with duplicate `user_pref()` calls with the same first argument. The later calls overwrite the earlier ones. This means we can just append to the file the default prefs we want to simplify the code we maintain here.

bin/run-utils.js
```diff
@@ -184,3 +184,195 @@ function runFirefox(options) {
...
+    });
+}
+
+exports.setUserPrefs = setUserPrefs;
+
+function restoreUserPrefs(profilePath) {
+  return writeUserPrefs(profilePath);
+}
+
+exports.restoreUserPrefs = restoreUserPrefs;
```
Mardak, line 319: This is unused?

package.json
```diff
@@ -34,7 +34,7 @@
     "express": "4.13.3",
     "firefox-profile": "0.3.11",
     "fs-promise": "0.3.1",
-    "fx-runner": "0.0.10",
+    "fx-runner": "1.0.0",
```
Mardak, line 37: Nice quick turnaround!

Mardak:
Also would be good to update `README.md` at the top currently with: **One time only**: This will use a new profile created in a `./loop-dev/`

Standard8:
> Also would be good to update `README.md` at the top currently with: One time only: This will use a new profile created in a `./loop-dev/`

Its `docs/Developing.md` now. Please update that as well. I think it'd be nice to explain how to create a new profile & how to continue using an existing one (if it hasn't been used for Loop before).

Also, if `loop-dev` is going to be the default, please can we add that to `.gitignore` so it doesn't get accidentally committed.

Thanks.

Standard8:

PR #62, opened 2016-01-08
Bug 1238170 - Make dev server room routing code more readable

dmose:
This makes the regexp more readable/maintainable and fixes the problem.

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -74,7 +74,14 @@ function serveIndex(req, res) {
 //
 app.get("/config.js", getConfigFile);
 
-app.get(/^\/[\w\-]+$/, serveIndex);
+var roomRouteRegexp = new RegExp(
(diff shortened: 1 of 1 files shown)


bin/server.js
```diff
@@ -74,7 +74,14 @@ function serveIndex(req, res) {
 //
 app.get("/config.js", getConfigFile);
 
-app.get(/^\/[\w\-]+$/, serveIndex);
+var roomRouteRegexp = new RegExp(
+  "^\\/" + // starts with a /
+  "[" +
+    "\\w" + // non-whitespace characters
+    "\\-" + // - used in the room query sometimes (for crypto?)
```
Standard8: I think you can get room keys with "-" as well, but yes, its generally for the crypto stuff.

PR #69, opened 2016-01-12
Bug 1236336 - pages without title show blankspace instead of null

kingroryg:
fix for this bug : https://bugzilla.mozilla.org/show_bug.cgi?id=1236336

Screenshot : 

screen shot 2016-01-12 at 8 36 36 pm

### DIFF
--- a/add-on/panels/js/roomViews.jsx
+++ b/add-on/panels/js/roomViews.jsx
@@ -521,6 +521,9 @@ loop.roomViews = (function(mozL10n) {
         var roomTitle = this.state.roomName ||
                         this.state.roomContextUrls[0].description ||
                         this.state.roomContextUrls[0].location;
+        if (typeof roomTitle === "object" && roomTitle === null) {
+          roomTitle = " ";
+        }
         this.setTitle(roomTitle);
       }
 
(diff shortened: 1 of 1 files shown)


add-on/panels/js/roomViews.jsx
```diff
@@ -521,6 +521,9 @@ loop.roomViews = (function(mozL10n) {
         var roomTitle = this.state.roomName ||
                         this.state.roomContextUrls[0].description ||
                         this.state.roomContextUrls[0].location;
+        if (typeof roomTitle === "object" && roomTitle === null) {
+          roomTitle = " ";
```
Standard8: I think we can change the if statement here to just check `!roomTitle`. There's no need to be more specific.

Also, please can you change what it sets the replacement title to:

```
roomTitle = mozL10n.get("room_name_untitled_page");
```

That should display "Untitled Page" like the room list in the panel. I think that'll be slightly more consistent.

kingroryg:
Made the required changes .

Standard8:
Great, this looks good. Please could you squash the two changesets into one, and change the commit message to:

"Bug 1236336 - If there is no page details, fallback to displaying untitled page rather than null. r=Standard8"

If you could also rebase it on latest master at the same time, that would be useful though not important.

kingroryg:
So should i submit a new PR ?

PR #71, opened 2016-01-12
Bug 1238509 - Drop some obsolete preferences

martinpinto:
Removed all unused references of loop.textChat.enabled and loop.learnMoreUrl.


add-on/panels/js/conversation.jsx
```diff
@@ -105,7 +105,6 @@ loop.conversation = (function(mozL10n) {
       ["GetAllStrings"],
       ["GetLocale"],
       ["GetLoopPref", "ot.guid"],
-      ["GetLoopPref", "textChat.enabled"],
```
Standard8, line 108: This isn't quite going to work as you might expect. The loop.requestMulti call below uses this to request the values from the browser. The results are returned in an ordered Array. In the `then` section, the array is iterated over and the values extracted.

So we actually need to remove the 

```
  var useDataChannels = results[++requestIdx];
```

as well.

I think for now, I'd like to keep the `useDataChannels` option to `loop.OTSdkDriver` available, so you can just set that to `true`.

Standard8:
This is a good start, I've just commented on the issue that I'd hadn't thought of previously.

If you could also rebase on latest master and force push your branch, that will fix the error that travis-ci is seeing - we had a minor issue with package versions earlier tonight that we've now fixed.

Thank you.

add-on/panels/js/conversation.jsx
martinpinto, line 108: Awesome I will work on this. Thank you!

add-on/panels/js/conversation.jsx
```diff
@@ -150,14 +149,11 @@ loop.conversation = (function(mozL10n) {
...
       });
 
-      // We want data channels only if the text chat preference is enabled.
-      var useDataChannels = results[++requestIdx];
 
       var dispatcher = new loop.Dispatcher();
       var sdkDriver = new loop.OTSdkDriver({
         constants: constants,
         isDesktop: true,
-        useDataChannels: useDataChannels,
```
Standard8, line 160: You need to keep this line but change it to:

```
useDataChannels: true
```

Otherwise the text chat doesn't get enabled.

PR #77, opened 2016-01-13
Bug 1232707 - Add commands to make a dist directory for exporting to mozilla-central and a basic export set.

Standard8:


Makefile
```diff
@@ -282,6 +283,24 @@ dist_xpi: add-on_dist
...
+	rm -f $(DIST_EXPORT_DIR)/install.rdf
+	$(RSYNC) $(ADD-ON)/install.rdf.in $(DIST_EXPORT_DIR)
+	# jar.mn is used for Firefox build
+	rm -f $(DIST_EXPORT_DIR)/chrome.manifest
+	# Add the jar file.
+	$(RSYNC) $(ADD-ON)/jar.mn $(DIST_EXPORT_DIR)
+	# Use the original prefs.js.
+	$(RSYNC) $(ADD-ON)/preferences/prefs.js $(DIST_EXPORT_DIR)/chrome/content/preferences/prefs.js
+	# Copy the vendor files that support the unit tests.
+	$(RSYNC) $(BUILT)/test/vendor/* $(DIST_EXPORT_DIR)/chrome/content/shared/test/vendor
```
mikedeboer, line 301: Why would the unit test vendor files be in the dist bundle? Why would the unit tests be there at all?
To me, it's the same as not having the .jsx files in there: less confusion.

Makefile
```diff
@@ -307,6 +326,30 @@ karma: build
 .PHONY: build
 build: add-on standalone ui vendor_libs
 
+GIT_EXPORT_LOCATION := ../gecko-dev
+GIT_EXPORT_DIR := $(GIT_EXPORT_LOCATION)/browser/extensions/loop
+
+.PHONY: git-export
```
mikedeboer, line 335: What's a 'git-export'?

Makefile
Standard8, line 301: This is for export to mozilla-central. If we don't export unit tests, then we can't have them running against the firefox builds.

The add-on dist goes to a separate location.

Makefile
Standard8, line 335: Exporting to a git based mozilla-central, although I haven't hooked up any git commands yet, so we could possibly rename it.

add-on/jar.mn
```diff
@@ -27,137 +27,21 @@
...
-  content/shared/js/validate.js                    (content/shared/js/validate.js)
-
-  # Shared libs
-#ifdef DEBUG
-  content/shared/vendor/react.js                   (content/shared/vendor/react.js)
-#else
-  content/shared/vendor/react.js                   (content/shared/vendor/react-prod.js)
+  locale/                        (chrome/locale/*)
+  skin/                          (chrome/skin/*)
+  content/modules/               (chrome/content/modules/*)
```
mikedeboer, line 32: Wildcards! Me gusta!

add-on/panels/vendor/l10n.js
```diff
@@ -221,14 +221,19 @@
...
+      },
+      get code() {
+        return gLanguage;
+      },
+      get direction() {
+        // http://www.w3.org/International/questions/qa-scripts
+        // Arabic, Hebrew, Farsi, Pashto, Urdu
+        var rtlList = ['ar', 'he', 'fa', 'ps', 'ur'];
+        return (rtlList.indexOf(gLanguage) >= 0) ? 'rtl' : 'ltr';
+      }
```
mikedeboer, line 236: I really dig these l10n changes. r=me for those! (Including the unit test updates)

One thing however, is that at the top of this file we note that this source came form pdf.js originally. Should we make an effort to keep these in sync, or can we obsolete that statement?

Standard8, line 236: We say its modified at the moment, I've added a note about matching to the API.

shared/test/frontend_tester.py
```diff
@@ -17,14 +17,59 @@
 # need to consider whether this code wants to be shared with WebDriver tests
 # for other browsers, though.
 #
+
+gCommonDir = None
```
dmose, line 24: Adding a comment explaining where this needs to be overridden and why would be nice.

dmose, line 24: And, really, what it represents (i.e. common between what and what other thing?).

shared/test/frontend_tester.py
```diff
@@ -17,14 +17,59 @@
...
+
+# These redirects map the paths expected by the index.html files to the paths
+# in mozilla-central. In the github repo, the unit tests are run entirely within
+# karma, where the files are loaded directly. For mozilla-central we must map
+# the files to the appropriate place.
+REDIRECTIONS = {
+    "/test/vendor": "/chrome/content/shared/test/vendor",
```
dmose, line 31: If there's some way to do avoid more hardcoded paths, since that bit us in the past, it _might_ be worth fixing or spinning off a new bug for that.  Or maybe we just don't worry about unless it becomes an issue.

shared/test/test_shared_all.py
```diff
@@ -10,7 +10,7 @@ class TestSharedUnits(BaseTestFrontendUnits):
 
     def setUp(self):
         super(TestSharedUnits, self).setUp()
-        self.set_server_prefix(".")
+        self.set_server_prefix("../../../../")
```
dmose, line 15: A comment about what this default assumes or is going to resolved against would be nice; similar suggestion for the other use.

shared/test/frontend_tester.py
```diff
@@ -17,14 +17,59 @@
...
+}
+
+
 class ThreadingSimpleServer(SocketServer.ThreadingMixIn,
                             BaseHTTPServer.HTTPServer):
     pass
 
 
-class QuietHttpRequestHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
+class HttpRequestHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
```
dmose, line 50: A Python doc string for the class as a whole would be nice.  Doc strings for the individual methods probably aren't necessary.

Makefile
```diff
@@ -307,6 +329,30 @@ karma: build
...
+git-export: build dist_export
+	find -E $(GIT_EXPORT_DIR) -type f ! -regex \
+	  '.*/(moz.build|README.txt|.gitignore|run-all-loop-tests.sh|manifest.ini)' -delete
+	$(RSYNC) $(DIST_EXPORT_DIR)/* $(GIT_EXPORT_DIR)
+	# TEMPORARY ITEMS to get l10n to work the "traditional" way in m-c whilst we
+	# sort out a proper fix.
```
dmose: Referencing a bug number here would be good.

Makefile
```diff
@@ -307,6 +329,30 @@ karma: build
...
+	# Drop locales from the jar file.
+	grep -v 'locale' $(ADD-ON)/jar.mn > $(GIT_EXPORT_DIR)/jar.mn
+	# Don't load our copy of loop.properties, load the browser's one instead.
```
dmose, line 344: It's not obvious to me why this is necessary.... is it so that a beta add-on can be installed at the same time as (and correctly override) the system add-on?  If so, that's worth a comment.

Or is it just part of the way the export ended up needing to be structured?

Standard8, line 344: This is just part of the workaround for the l10n issues.

PR #80, opened 2016-01-14
Bug 1239783 - NewTabURL.override is deprecated so use aboutNewTabService.newTabURL instead

Mardak:


dmose:
Ah, good point.  Reviewing now...

dmose:
Looks good to me; r=@dmose

Mardak:
The deprecation warning is removed by bug 1239085, so we'll keep using the NewTabURL API made for add-ons.

PR #81, opened 2016-01-14
Bug 1239780 - Port Bug 1101817 - Part 5: Remove WeakMap.prototype.clear from Loop addon. Also account for different Firefox versions.

Standard8:
This fixes the add-on running against latest nightly - Bug 1101817 changed the API, so we need to detect that and select the call appropriately.


Mardak:
r=Mardak

PR #84, opened 2016-01-14
Bug 1239634 - Change some new user journey strings

chrafuse:


locale/en-US/add-on.properties
```diff
@@ -183,14 +184,17 @@ door_hanger_button=OK
 
 infobar_screenshare_browser_message2=You are sharing your tabs. Any tab you click on can be seen by your friends
 infobar_screenshare_paused_browser_message=Tab sharing is paused
+infobar_screenshare_stop_sharing_message=You are no longer sharing your tabs
 infobar_button_pause_label=Pause
 infobar_button_pause_accesskey=P
-infobar_button_restart_label=Restart
+infobar_button_restart_label2=Restart sharing
```
Mardak, line 186: Is "Restart" supposed to replace "Resume"? If so, we should be able to remove the Resume label/accesskey and use "R" for Restart? Also, this new label should probably be used in code somewhere.

locale/en-US/add-on.properties
```diff
@@ -109,6 +109,7 @@ incoming_call_title2=Conversation Request
 incoming_call_block_button=Block
 hangup_button_title=Hang up
 hangup_button_caption2=Exit
+hangup_message=You are no longer sharing your tabs
```
Mardak: Where is this string supposed to be used? The name of the string `hangup_message` doesn't seem to match the string.

locale/en-US/add-on.properties
```diff
@@ -183,14 +184,17 @@ door_hanger_button=OK
...
-infobar_button_restart_label=Restart
+infobar_button_restart_label2=Restart sharing
 infobar_button_restart_accesskey=e
 infobar_button_resume_label=Resume
 infobar_button_resume_accesskey=R
-infobar_button_stop_label=Stop
+infobar_button_stop_label2=Stop sharing
 infobar_button_stop_accesskey=S
+infobar_button_disconnect_label=Disconnect
+infobar_button_disconnect_accesskey=X
```
Mardak: Where did you get the `X` accesskey for disconnect?

chrafuse: Changed to D from X

chrafuse:
Removed pause label and key and replaced it with Stop sharing label and key.
Changes per UX.

chrafuse:
Buttons now adjust length to the string given.

add-on/chrome/bootstrap.js
```diff
@@ -534,13 +534,13 @@ var WindowListener = {
...
+                                               "infobar_button_stop_label2");
         let pauseButtonAccessKey = this._getString(this._browserSharePaused ?
-                                                   "infobar_button_resume_accesskey" :
-                                                   "infobar_button_pause_accesskey");
+                                                   "infobar_button_restart_accesskey" :
+                                                   "infobar_button_stop_accesskey");
         let barLabel = this._getString(this._browserSharePaused ?
-                                       "infobar_screenshare_paused_browser_message" :
+                                       "infobar_screenshare_stop_sharing_message" :
                                        "infobar_screenshare_browser_message2");
```
Mardak: This duplication seems a bit silly. Perhaps a simple helper can clean it up.

```
let getPausedUnpausedStrings = () => this._browserSharePaused ? {
  label: this._getString("infobar_button_restart_label2"),
  accesskey: this._getString("infobar_button_restart_accesskey"),
  message: this._getString("infobar_screenshare_stop_sharing_message")
} : {
  ...
};
```

Then call it outside and inside the callback then select the appropriate key in desired places.

add-on/chrome/skin/shared/loop.css
```diff
@@ -180,11 +180,16 @@
...
+
   notification[value="loop-sharing-notification"] .notification-button {
     -moz-appearance: none;
     background-color: #fff;
     border: 0;
-    width: 100px;
+    padding: .2rem 1rem .4rem;
     height: 40px;
     margin: 0;
     list-style-image: url(chrome://loop/content/shared/img/pause-12x12.svg);
```
Mardak: You didn't touch this line, but do we actually want this list-style-image here? If not, we should remove it.

add-on/chrome/skin/shared/loop.css
```diff
@@ -180,11 +180,16 @@
...
+    padding-top: .2rem;
+    padding-bottom: .4rem;
+  }
+
   notification[value="loop-sharing-notification"] .notification-button {
     -moz-appearance: none;
     background-color: #fff;
     border: 0;
-    width: 100px;
+    padding: .2rem 1rem .4rem;
```
Mardak, line 192: Nice! Assuming these are the spacing we want.

locale/en-US/add-on.properties
```diff
@@ -26,7 +26,7 @@ sign_in_again_button=Sign In
 sign_in_again_use_as_guest_button2=Use {{clientSuperShortname}} as a Guest
 
 panel_browse_with_friend_button=Browse this page with a friend
-panel_stop_sharing_tabs_button=Stop sharing your tabs
+panel_stop_sharing_tabs_button2=Disconnect
```
Mardak: The string key should probably just be `panel_disconnect_button`

chrafuse:
Changed disconnect label name.
Added getPausedUnpausedStrings function.

add-on/chrome/bootstrap.js
```diff
@@ -533,15 +533,22 @@ var WindowListener = {
...
+          label: this._getString("infobar_button_stop_label2"),
+          accesskey: this._getString("infobar_button_stop_accesskey"),
+          message: this._getString("infobar_screenshare_browser_message2")
+        };
+        let stringObj = getPausedUnpausedStrings();
+
+        let pauseButtonLabel = stringObj.label;
+        let pauseButtonAccessKey = stringObj.accessKey;
+        let barLabel = stringObj.message;
+
```
Mardak: These 3 `let` declarations are unnecessary. Just directly use `stringObj` in the various places below.

chrafuse:
Added a commit that may be better use of preloaded strings instead of calling getString every time. I will remove the above unnecessary declarations and push.

chrafuse:
Removed unnecessary lets and based on pre-load string commit.

add-on/chrome/bootstrap.js
```diff
@@ -533,45 +533,44 @@ var WindowListener = {
...
+          pausedStrings.message,
           kBrowserSharingNotificationId,
           // Icon is defined in browser theme CSS.
           null,
           box.PRIORITY_WARNING_LOW,
           [{
-            label: pauseButtonLabel,
-            accessKey: pauseButtonAccessKey,
+            label: pausedStrings.label,
+            accessKey: pausedStrings.accessKey,
```
Mardak: I believe the infobar gets reshown when you switch tabs even when you're paused. Double check that pausing then switching tabs keeps the text showing the paused state.

chrafuse: Seems to hold the state correctly on new and switched tabs.

Mardak: Something isn't quite right. I just checked out the branch and I'm seeing "[pause icon] Restart sharing"

chrafuse:
Combined commits into 1 commit.  Corrected initial strings to check if paused or not. Tested manually.

PR #85, opened 2016-01-15
Bug 1196143 - IDN homograph attack in firefox hello chat

chrafuse:
use of shared.Utils.formatURL method.

### DIFF
--- a/shared/js/linkifiedTextView.jsx
+++ b/shared/js/linkifiedTextView.jsx
@@ -75,7 +75,7 @@ loop.shared.views.LinkifiedTextView = (function() {
       var elements = [];
       var result = loop.shared.urlRegExps.fullUrlMatch.exec(s);
       var reactElementsCounter = 0; // For giving keys to each ReactElement.
-
+      var sanitizeURL;
       while (result) {
         // If there's text preceding the first link, push it onto the array
         // and update the string pointer.
@@ -85,10 +85,12 @@ loop.shared.views.LinkifiedTextView = (function() {
         }
 
         // Push the first link itself, and advance the string pointer again.
+        // Bug 1196143 - formatURL sanitizes(decodes) the URL from IDN homographic attacks.
+        sanitizeURL = loop.shared.utils.formatURL(result[0]).location;
         elements.push(
-          <a { ...this._generateLinkAttributes(result[0]) }
+          <a { ...this._generateLinkAttributes(sanitizeURL) }
             key={reactElementsCounter++}>
-            {result[0]}
+            {sanitizeURL}
           </a>
         );
         s = s.substr(result[0].length);
--- a/shared/js/utils.js
+++ b/shared/js/utils.js
@@ -367,20 +367,29 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
   /**
    * Formats a url for display purposes. This includes converting the
    * domain to punycode, and then decoding the url.
+   * Intended to be used for both display and (uglier in confusing cases) clickthrough,
+   * as described by dveditz in comment 12 of the bug 1196143,
+   * as well as testing the behavior case in the browser.
    *
-   * @param {String} url The url to format.
-   * @return {Object}    An object containing the hostname and full location.
+   * @param {String}  url                   The url to format.
+   * @param {String}  suppressConsoleError  For testing, call with a boolean which is true to squash the default console error.
+   * @return {Object}                       An object containing the hostname and full location.
    */
-  function formatURL(url) {
+  function formatURL(url, suppressConsoleError) {
     // We're using new URL to pass this through the browser's ACE/punycode
     // processing system. If the browser considers a url to need to be
     // punycode encoded for it to be displayed, then new URL will do that for
     // us. This saves us needing our own punycode library.
+    // Note that URL does canonicalize hostname-only URLs,
+    // adding a slash to them, but this is ok for at least HTTP(S)
+    // because GET always has to specify a path, which will (by default) be
     var urlObject;
     try {
       urlObject = new URL(url);
     } catch (ex) {
-      console.error("Error occurred whilst parsing URL:", ex);
+      if (suppressConsoleError ? !suppressConsoleError : true) {
+        console.error("Error occurred whilst parsing URL:", ex);
+      }
       return null;
     }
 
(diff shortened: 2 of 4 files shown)


shared/js/utils.js
```diff
@@ -368,19 +368,24 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
...
+   * @return {Object}         An object containing the hostname and full location.
    */
-  function formatURL(url) {
+  function formatURL(url, noError) {
     // We're using new URL to pass this through the browser's ACE/punycode
     // processing system. If the browser considers a url to need to be
     // punycode encoded for it to be displayed, then new URL will do that for
     // us. This saves us needing our own punycode library.
+    // URL is canonicalizing hostname-only URLs, as GET always has to include a path
+    // and correctly adds a backslash.
```
dmose: Maybe "Note that URL does canonicalize hostname-only URLs, adding a slash to them, but this is ok for at least HTTP(S) because GET always has to specify a path, which will (by default) be /"

@adamroach Is it reasonable to assume that this is ok for non-HTTP(S) URLs as well?

shared/js/linkifiedTextView.jsx
```diff
@@ -85,10 +85,11 @@ loop.shared.views.LinkifiedTextView = (function() {
         }
 
         // Push the first link itself, and advance the string pointer again.
+        safeURL = loop.shared.utils.formatURL(result[0]).location;
```
dmose: Please add a comment explaining why it's important to call this here.

shared/js/utils.js
```diff
@@ -368,19 +368,24 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
    * Formats a url for display purposes. This includes converting the
    * domain to punycode, and then decoding the url.
```
dmose, line 369: Let's make this comment clear that these are intended to be used for both display and (uglier in confusing cases) clickthrough, as described by dveditz in comment 12 of the bug, as well as testing the behavior case in the browser.

shared/js/linkifiedTextView.jsx
dmose: Feel free to also change the name to sanitizeURL

shared/js/utils.js
```diff
@@ -368,19 +368,24 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
...
    *
-   * @param {String} url The url to format.
-   * @return {Object}    An object containing the hostname and full location.
+   * @param {String}  url     The url to format.
+   * @param {String}  noError Called with a boolean which is true to squash the console error.
+   * @return {Object}         An object containing the hostname and full location.
    */
-  function formatURL(url) {
+  function formatURL(url, noError) {
```
dmose: Let's switch this to "suppressConsoleError".

dmose: Or kill this, and...

shared/js/utils.js
```diff
@@ -368,19 +368,24 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
...
     // punycode encoded for it to be displayed, then new URL will do that for
     // us. This saves us needing our own punycode library.
+    // URL is canonicalizing hostname-only URLs, as GET always has to include a path
+    // and correctly adds a backslash.
     var urlObject;
     try {
       urlObject = new URL(url);
     } catch (ex) {
-      console.error("Error occurred whilst parsing URL:", ex);
+      if (noError ? !noError : true) {
```
dmose: Switch this to console.log instead of console.error.

shared/js/views.jsx
```diff
@@ -547,7 +547,7 @@ loop.shared.views = (function(_, mozL10n) {
       var hostname;
 
       try {
-        hostname = new URL(this.props.url).hostname;
+        hostname = loop.shared.utils.formatURL(this.props.url, true).hostname;
```
dmose: Since we have a try-catch inside formatURL, we shouldn't need it here.

shared/test/linkifiedTextView_test.js
```diff
@@ -197,24 +197,29 @@ describe("loop.shared.views.LinkifiedTextView", function() {
           markup: '<p><a href="http://example.com/">http://example.com/</a></p>'
         },
         {
+          desc: "should linkify a string containing only a URL with no trailing slash",
+          rawText: "http://example.com",
+          markup: '<p><a href="http://example.com/">http://example.com/</a></p>'
+        },
+        {
```
dmose, line 204: Please check to see if it's redundant, and if so, remove this test.

dmose:
Looks good with noted changes; I'd like to see it once more before landing.

shared/test/linkifiedTextView_test.js
chrafuse, line 204: Not redundant, tests output of http://example.com and the previous test tests http://example.com/
Making sure it adds the slash and that it does not add extra slashes..

shared/js/views.jsx
chrafuse: had to leave in as it returns null for invalid url we could have used an if statement also.

chrafuse:
@dmose added comments and kept the try catch due to return null and not an error message for invalid urls.  (ie. http://kfdhfkdshfdsahf).  I will push this master if all good.

shared/js/utils.js
```diff
@@ -367,20 +367,29 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
...
    * Formats a url for display purposes. This includes converting the
    * domain to punycode, and then decoding the url.
+   * Intended to be used for both display and (uglier in confusing cases) clickthrough,
+   * as described by dveditz in comment 12 of the bug 1196143,
+   * as well as testing the behavior case in the browser.
    *
-   * @param {String} url The url to format.
-   * @return {Object}    An object containing the hostname and full location.
+   * @param {String}  url                   The url to format.
+   * @param {String}  suppressConsoleError  Called with a boolean which is true to squash the console error.
```
dmose: Please add something like (for testing) here so it's clear why we have this.

shared/test/views_test.js
```diff
@@ -444,7 +444,6 @@ describe("loop.shared.views", function() {
       view = mountTestComponent({
         url: "fjrTykyw"
       });
-
```
dmose, line 447: Please revert this; blank lines are used to separate the setup/execute/verify steps of a unit test.

dmose:
Looks good, r=dmose once you've fixed the two mentioned nits, go ahead and land.

PR #86, opened 2016-01-15
Bug 1239963 - Stop unloading style sheets and destorying the loop-button on shutdown to help performance and Talos test failures.

Standard8:


add-on/chrome/bootstrap.js
```diff
@@ -875,6 +875,12 @@ function shutdown() {
   // Stop waiting for browser windows to open.
   wm.removeListener(WindowListener);
 
+  // If the app is shutting down, don't worry about cleaning up, just let
+  // it fade away...
+  if (reason == 2) {
```
Mardak: You can use the magic global `APP_SHUTDOWN` instead of `2` (and update add-on/chrome/.eslintrc)

mikedeboer: +1. r=me with that change.

Standard8: Oooh, I didn't realise that, much nicer!

PR #87, opened 2016-01-15
Bug 1236336 - If there is no page details, fallback to displaying untitled page rather than null. r=Standard8

kingroryg:


Standard8:
Hi @sar95. You need to squash the two commits into one. You should be able to do something like:
- `git rebase -i upstream/master`
- In the editor that appears, change the second line so that `pick` is replaced by `f`
- That should make the two commits into one.
- If you then do `git commit --amend -m "Bug 1236336 - If there is no page details, fallback to displaying untitled page rather than null. r=Standard8"` that will change the commit message
- You can then `git push -f origin master` to push the updated changes.

These instructions are slightly special for your case as you're currently working off master - its generally easier to keep master following upstream/master and do you work on branches. There's various resources about how to do this for future patches, but I'm happy to help if you get stuck.

kingroryg:
screen shot 2016-01-19 at 10 24 35 pm

This is what I get after running the first command . Second line is empty. Do I replace noop ?

Standard8:
Hmm, try a `git fetch upstream` first. That might work better.

kingroryg:
I get the same thing in editor .  Sorry, for the trouble here .  I think I should close this , refork and submit a fresh PR . Would be better ?

Standard8:
Ok, I'm not quite sure what's going on here. Do you mind if I import your patches and push them from my local branch? I'll include you as author, of course.

kingroryg:
Yeah , sure . No problem . This was a learning experience . :)

Standard8:
Landed as https://github.com/mozilla/loop/commit/f6755be5b8983c6862efc63403575fbc73f89991

PR #90, opened 2016-01-19
Bug 1210588 - share pointer from desktop to standalone

mancas:


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

add-on/chrome/bootstrap.js
```diff
@@ -485,6 +488,9 @@ var WindowListener = {
           // metadata about the page is available when this event fires.
           gBrowser.addEventListener("DOMTitleChanged", this);
           this._browserSharePaused = false;
+
+          this._mousemoveHandler = this.handleMousemove.bind(this);
```
fcampo: I think we can avoid `_mousemoveHandler` and just assign `handleMousemove` to the event listener. 
What's the use of keeping the second variable?

add-on/chrome/bootstrap.js
```diff
@@ -646,6 +654,31 @@ var WindowListener = {
...
+        if (deltaX < 0 || deltaX > browserBox.width ||
+            deltaY < 0 || deltaY > browserBox.height ||
+            (Math.abs(deltaX - this.lastCursorX) < MIN_CURSOR_DELTA &&
+             Math.abs(deltaY - this.lastCursorY) < MIN_CURSOR_DELTA)) {
+          return;
+        }
+        this.lastCursorX = deltaX;
+        this.lastCursorY = deltaY;
+
+        this.LoopAPI.broadcastPushMessage("CursorPosition",
```
fcampo: nit: as it is an event, an action, I'd name it `CursorPositionChange`

shared/js/activeRoomStore.js
```diff
@@ -968,6 +996,7 @@ loop.store.ActiveRoomStore = (function() {
       loop.request("AddBrowserSharingListener", this.getStoreState().windowId)
         .then(this._browserSharingListener);
       loop.subscribe("BrowserSwitch", this._browserSharingListener);
+      loop.subscribe("CursorPosition", this._sendCursorPositionListener);
```
fcampo: as in comment https://github.com/mozilla/loop/pull/90/files#r50117129, `CursorPositionChange`

shared/js/otSdkDriver.js
```diff
@@ -776,6 +789,21 @@ loop.OTSdkDriver = (function() {
     },
 
     /**
+     * Sends the cursor position on the data channel.
+     *
+     * @param {String} message The message to send.
+     */
+    sendCursorMessage: function(message) {
+      if (!this._publisherChannel || !this._subscriberChannel) {
+        return;
```
fcampo, line 857: maybe add a `console.error` if no channel?

fcampo, line 857: also, why checking for `_subscriberChannel` if we don't use it later? is it the default if primary fails?

shared/js/utils.js
```diff
@@ -109,6 +109,10 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
     CONTEXT_TILE: "context-tile"
   };
 
+  var CURSOR_MESSAGE_TYPES = {
```
fcampo, line 112: :+1:

shared/js/views.jsx
```diff
@@ -599,22 +599,94 @@ loop.shared.views = (function(_, mozL10n) {
       isLoading: React.PropTypes.bool.isRequired,
       mediaType: React.PropTypes.string.isRequired,
       posterUrl: React.PropTypes.string,
+      remoteCursorLeft: React.PropTypes.number,
```
fcampo: if we're always sending/receiving this in pairs, why not using a 

``` javascript
remoteCursorPosition: {
  left: number,
  top: number
}
```

shared/js/views.jsx
```diff
@@ -599,22 +599,94 @@ loop.shared.views = (function(_, mozL10n) {
...
+
     componentDidMount: function() {
       if (!this.props.displayAvatar) {
         this.attachVideo(this.props.srcMediaElement);
       }
     },
 
+    componentWillUnmount: function() {
+      var videoElement = this.getDOMNode().querySelector("video");
+      videoElement.removeEventListener("timeupdate", this.handleVideoUpdate);
```
fcampo: where is this listener added and why is it needed to be removed here?

shared/js/views.jsx
```diff
@@ -831,6 +927,32 @@ loop.shared.views = (function(_, mozL10n) {
     }
   });
 
+  var RemoteCursorView = React.createClass({
+    mixins: [React.addons.PureRenderMixin],
+
+    propTypes: {
+      offsetLeft: React.PropTypes.number,
```
fcampo: all the props are always required?

shared/js/views.jsx
```diff
@@ -831,6 +927,32 @@ loop.shared.views = (function(_, mozL10n) {
...
+      remoteCursorLeft: React.PropTypes.number,
+      remoteCursorTop: React.PropTypes.number
+    },
+
+    render: function () {
+      console.log("remoteCursorTop", this.props.remoteCursorTop);
+      console.log("remoteCursorLeft", this.props.remoteCursorLeft);
+      console.log("offsetTop", this.props.offsetTop);
+      console.log("offsetLeft", this.props.offsetLeft);
+      var cursorStyle = {
```
fcampo: style, not position?

add-on/chrome/bootstrap.js
mancas: I'm using the variable just because we need to remove the handler when the user leave the room, or stops the tab sharing (see line 514)

shared/js/otSdkDriver.js
mancas, line 857: We must ensure that both users are ready to receive and send information. If an user has entered in the room but hasn't subscribed to the video stream yet, we should prevent sending it the cursor information because it will display while the stream video is still loading.

shared/js/views.jsx
mancas: We add the listener in |attachVideo| and this method it's just to cleanup the listeners before react remove the element from the dom. Perhaps we can rid off it, wdyt?

shared/js/views.jsx
mancas: Yeah, this variable defines the cursor position but i keep that name because we're setting the style of the element

shared/js/views.jsx
mancas: you mean we should set them as required? In fact, all of them are necessary to set a correct position

shared/js/otSdkDriver.js
Standard8, line 857: I was going to suggest checking that the publisher channel was subscribed to at the other end, unfortunately the SDK doesn't give us events to know that. So I think checking for publisher & subscriber makes sense. A comment in the code would be useful.

Standard8, line 857: Also, I'm not sure if you're going to address it here or not, but we need to make sure that we create a separate data channel for the pointer sharing.

This means we won't flood the text chat channel, and we can also likely make the pointer data channel lossy (we'll need to investigate if we're able to supply parameters to it and what they should be).

shared/js/activeRoomStore.js
```diff
@@ -946,6 +963,18 @@ loop.store.ActiveRoomStore = (function() {
       }.bind(this));
     },
 
+    _handleSendCursorPosition: function(event) {
+      var clientWidth = window.outerWidth;
+      var clientHeight = window.outerHeight;
```
Mardak: This looks a little dangerous to me primarily that I'm not entirely sure what `window` is referring to in this context. I think it would be safer to have the mousemove event listener calculate the percentage as it knows about gBrowser, tabs, etc.

shared/js/otSdkDriver.js
```diff
@@ -754,6 +744,29 @@ loop.OTSdkDriver = (function() {
       }.bind(this));
     },
 
+    _handleMessage: function(ev) {
+      try {
+        var message = JSON.parse(ev.data);
+        /* Append the timestamp. This is the time that gets shown. */
+        message.receivedTimestamp = (new Date()).toISOString();
+
+        if (message.type) {
```
Mardak: I suppose in implementing @Standard8's suggestion of a separate 'cursor' channel, we wouldn't need to check if a type exists when demuxing a message as 'text' will be exclusively chat messages and 'cursor' would be of position or click, etc.

shared/js/views.jsx
```diff
@@ -599,22 +599,100 @@ loop.shared.views = (function(_, mozL10n) {
...
+        streamVideoHeight = realVideoHeight;
+      }
+
+      this.setState({
+        cursorOffset: {
+          left: (clientWidth - streamVideoWidth) / 2,
+          top: (clientHeight - streamVideoHeight) / 2
+        },
+        streamVideoHeight: streamVideoHeight,
+        streamVideoWidth: streamVideoWidth
```
Mardak: Did you mean to use the state of `streamVideoHeight` and width somewhere else?

shared/js/views.jsx
```diff
@@ -831,6 +927,30 @@ loop.shared.views = (function(_, mozL10n) {
     }
   });
 
+  var RemoteCursorView = React.createClass({
+    mixins: [React.addons.PureRenderMixin],
+
+    propTypes: {
+      remoteCursorOffset: React.PropTypes.object,
+      remoteCursorPosition: React.PropTypes.object
```
Mardak: Does `RemoteCursorView` need to know about an offset and position? Any reason not to just have the props be a calculated top and left position?

shared/js/views.jsx
```diff
@@ -599,22 +599,100 @@ loop.shared.views = (function(_, mozL10n) {
...
       mediaType: React.PropTypes.string.isRequired,
       posterUrl: React.PropTypes.string,
+      remoteCursorPosition: React.PropTypes.object,
       // Expecting "local" or "remote".
       srcMediaElement: React.PropTypes.object
     },
 
+    getInitialState: function() {
+      return {
+        cursorOffset: {
```
Mardak: We happen to use the offset data for the cursor, but the values are actually referring to the video letterboxing, so the name should probably be describing "what it is" as opposed to the current one usage.

add-on/chrome/bootstrap.js
Mardak: Instead of binding, you could do something similar to line 489: `gBrowser.addEventListener("mousemove", this);` where the `mousemove` event will be passed to `this.handleEvent`, which already has a `switch (event.type)` to dispatch appropriately -- in this case, dispatch to `this.handleMousemove`.

add-on/chrome/bootstrap.js
```diff
@@ -646,6 +654,31 @@ var WindowListener = {
...
+          return;
+        }
+        this.lastCursorTime = now;
+
+        // Skip the update if cursor is out of bounds or didn't move much.
+        let browserBox = gBrowser.selectedBrowser.boxObject;
+        let deltaX = event.screenX - browserBox.screenX;
+        let deltaY = event.screenY - browserBox.screenY;
+        if (deltaX < 0 || deltaX > browserBox.width ||
+            deltaY < 0 || deltaY > browserBox.height ||
```
Mardak: As commented elsewhere about percentages, I think it would be safer to divide by `browserBox.width` instead of `window.outerWidth` as we're already using it to compute bounds.

shared/img/cursor.svg
```diff
@@ -0,0 +1,25 @@
...
+  </style>
+  <defs>
+    <g id="cursor">
+      <path d="M3.38697632,13.4830316 L6.22,18.492 L6.689,19.321 L7.54,18.893 L9.519,17.898 L9.518,17.898 L10.456,17.426 L9.939,16.512 L7.38622856,11.999 L12.63,11.999 L10.664,10.252 L1.664,2.252 L0,0.773 L0,3 L0,2.999 L0,17.999 L1.8,15.599 L3.38697632,13.4830316 L3.38697632,13.4830316 Z" id="Shape" fill="#FFFFFF"></path>
+      <path fill-rule="evenodd" d="M3.505,11.66 L7.091,18 L9.07,17.005 L5.673,11 L10,11 L1,3 L1,15 L3.505,11.66 L3.505,11.66 Z" id="Shape"></path>
+    </g>
+  </defs>
+  <use id="cursor-orange" xlink:href="#cursor"/>
+  <use id="cursor-blue" xlink:href="#cursor"/>
+</svg>
```
Mardak: We'll want to `svgo` to shrink this file.

shared/js/actions.js
```diff
@@ -131,6 +131,27 @@ loop.shared.actions = (function() {
     }),
 
     /**
+     * Used to send a message to the other peer.
+     */
+    SendCursorPosition: Action.define("sendCursorPosition", {
+      type: String,
```
Mardak: If we're including the `type` in the object, the name action should probably be more of `SendCursorData`

shared/js/views.jsx
Standard8: The other option is we move more of the positioning logic into `RemoteCursorView` - I'm not quite sure it makes sense, so feel free to ignore.

What I'm vaguely thinking about is you could pass it the video size and the values received for the actual remote cursor position, then have the `RemoteCursorView` do the necessary calculations for positioning it on the display.

That way, we keep the location calculations next to the item that is being positioned.

shared/js/activeRoomStore.js
```diff
@@ -857,6 +859,21 @@ loop.store.ActiveRoomStore = (function() {
         actionData.state === SCREEN_SHARE_STATES.ACTIVE);
     },
 
+    receivedCursorPosition: function(actionData) {
```
Standard8: We might want to move the new functionality in this file into a new store, similar to how the `TextChatStore` works.

One advantage of doing this, is that the `RemoteCursorView` could subscribe to this store for the remote cursor position location. Then as I said in my comment for that view, you could just pass in the video size via the props, and have the view calculate the actual position to display the cursor.

shared/test/index.html
```diff
@@ -57,6 +57,7 @@
   <script src="/shared/js/textChatView.js"></script>
   <script src="/shared/js/urlRegExps.js"></script>
   <script src="/shared/js/linkifiedTextView.js"></script>
+  <script src="/shared/js/remoteCursorStore.js"></script>
```
Mardak, line 60: You'll need to update karma.*.js files to have it load remoteCursorStore.js for add-on and standalone tests:

https://github.com/mozilla/loop/blob/master/test/karma/karma.coverage.desktop.js#L34
`+    "built/add-on/chrome/content/shared/js/remoteCursorStore.js",`

https://github.com/mozilla/loop/blob/master/test/karma/karma.coverage.shared_standalone.js#L35
`+    "built/standalone/content/shared/js/remoteCursorStore.js",`

shared/js/views.jsx
```diff
@@ -631,13 +663,17 @@ loop.shared.views = (function(_, mozL10n) {
...
+      var videoElement = this.getDOMNode().querySelector("video");
 
-      if (videoElement.tagName.toLowerCase() !== "video") {
+      if (!videoElement || videoElement.tagName.toLowerCase() !== "video") {
         // Must be displaying the avatar view, so don't try and attach video.
         return;
       }
 
+      if (this.props.shouldRenderRemoteCursor) {
+        videoElement.addEventListener("loadeddata", this.handleVideoUpdate);
```
Mardak: I believe this will only trigger once, so it will only record the initial `<video>` element size and the initial stream size. This results in wrong cursor positions if either the link clicker or link generator resizes the window. For the latter, I'm not sure if there's a handy "streamresized" event, so if not, we could send height and width as part of the cursor data.

shared/js/remoteCursorStore.js
```diff
@@ -0,0 +1,89 @@
...
+     * Receives cursor data.
+     *
+     * @param {sharedActions.receivedCursorData} actionData
+     */
+    receivedCursorData: function(actionData) {
+      switch (actionData.type) {
+        case CURSOR_MESSAGE_TYPES.POSITION:
+          var sent = new Date(actionData.sentTimestamp);
+          var received = new Date(actionData.receivedTimestamp);
+          console.log("Delay in ms", received.getTime() - sent.getTime());
```
Mardak: Do we need to keep the timestamps for the action? Or is this only for current debugging?

add-on/chrome/bootstrap.js
```diff
@@ -641,10 +647,43 @@ var WindowListener = {
...
+            (Math.abs(deltaX - this.lastCursorX) < MIN_CURSOR_DELTA &&
+             Math.abs(deltaY - this.lastCursorY) < MIN_CURSOR_DELTA)) {
+          return;
+        }
+        this.lastCursorX = deltaX;
+        this.lastCursorY = deltaY;
+
+        this.LoopAPI.broadcastPushMessage("CursorPositionChange", {
+          cursorX: (deltaX * 100) / browserBox.width,
+          cursorY: (deltaY * 100) / browserBox.height
```
Mardak: As noted for the issue of detecting whether the link generator has resized the window, instead of calculating a percent, we can just send (deltaX, deltaY, width, height) so the link clicker getting a cursor event immediately knows that the dimensions changed (and potentially has different letterboxing).

If we don't need to send the width or height; and we're sending a percent instead, let's just keep the number between 0 and 1 and call it something like `xRatio` or `ratioX`. `cursorX` will probably be easily confused for the absolute coordinates.

shared/css/common.css
```diff
@@ -591,3 +591,20 @@ html[dir="rtl"] .context-wrapper > .context-preview {
...
+}
+
+.remote-video-box > .remote-cursor {
+  width: 15px;
+  height: 20px;
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 65534;
+  background-color: red;
```
Mardak: Don't need the top/left/background-color styles. Please keep the rules alphabetical.

shared/js/views.jsx
```diff
@@ -599,22 +599,54 @@ loop.shared.views = (function(_, mozL10n) {
...
       }
     },
 
+    componentWillUnmount: function() {
+      var videoElement = this.getDOMNode().querySelector("video");
+      if (!this.props.shouldRenderRemoteCursor || !videoElement) {
+        return;
+      }
+
+      videoElement.removeEventListener("timeupdate", this.handleVideoUpdate);
```
Mardak: Did you mean to remove `"loadeddata"`? Although with `timeupdate`, you might be able to detect stream and element size changes?

add-on/chrome/bootstrap.js
```diff
@@ -485,6 +488,8 @@ var WindowListener = {
           // metadata about the page is available when this event fires.
           gBrowser.addEventListener("DOMTitleChanged", this);
           this._browserSharePaused = false;
+
+          gBrowser.addEventListener("mousemove", this);
```
Mardak, line 494: Add a comment about how this is added to the parent `gBrowser` to avoid adding/removing for each individual tab's browsers.

add-on/chrome/bootstrap.js
```diff
@@ -641,10 +647,43 @@ var WindowListener = {
...
             }
             break;
+          case "mousemove":
+            this.handleMousemove(event);
+            break;
           }
       },
 
       /**
+       * Handles mousemove events from gBrowser.
```
Mardak: A more descriptive comment about packaging up a cursor change message.

add-on/panels/js/conversation.jsx
```diff
@@ -188,10 +188,14 @@ loop.conversation = (function(mozL10n) {
...
       });
+      var remoteCursorStore = new loop.store.RemoteCursorStore(dispatcher, {
+        sdkDriver: sdkDriver
+      });
 
       loop.store.StoreMixin.register({
         conversationAppStore: conversationAppStore,
-        textChatStore: textChatStore
+        textChatStore: textChatStore,
+        remoteCursorStore: remoteCursorStore
```
Mardak: Move `remoteCursorStore` as the second/middle item to keep it alphabetical.

add-on/panels/panel.html
```diff
@@ -24,12 +24,13 @@
...
     <script type="text/javascript" src="shared/js/mixins.js"></script>
+    <script type="text/javascript" src="shared/js/store.js"></script>
+    <script type="text/javascript" src="shared/js/activeRoomStore.js"></script>
+    <script type="text/javascript" src="shared/js/remoteCursorStore.js"></script>
     <script type="text/javascript" src="shared/js/views.js"></script>
     <script type="text/javascript" src="shared/js/validate.js"></script>
     <script type="text/javascript" src="shared/js/actions.js"></script>
     <script type="text/javascript" src="shared/js/dispatcher.js"></script>
-    <script type="text/javascript" src="shared/js/store.js"></script>
-    <script type="text/javascript" src="shared/js/activeRoomStore.js"></script>
```
Mardak, line 32: Is moving store/activeRoomStore up necessary? Or this is just to keep those stores together? Having it above and together is fine.

shared/js/actions.js
```diff
@@ -131,6 +131,17 @@ loop.shared.actions = (function() {
     }),
 
     /**
+     * Notifies that a message has been received from the other peer.
```
Mardak: This comment isn't quite right.

shared/js/otSdkDriver.js
```diff
@@ -693,18 +694,7 @@ loop.OTSdkDriver = (function() {
...
-              /* Append the timestamp. This is the time that gets shown. */
-              message.receivedTimestamp = (new Date()).toISOString();
-
-              this.dispatcher.dispatch(
-                new sharedActions.ReceivedTextChatMessage(message));
-            } catch (ex) {
-              console.error("Failed to process incoming chat message", ex);
-            }
-          }.bind(this),
+          message: this._handleMessage.bind(this),
```
Mardak: Let's go ahead and handle the messages for a dedicated cursor channel, so we  can then leave this code for text chat messages untouched. The additional channel can be part of the 2nd commit.

Mardak:
You can ignore my comments about including the width/height in the cursor event. We should already be getting events/actions of the dimension changes:

https://github.com/mozilla/loop/blob/master/shared/js/otSdkDriver.js#L992

Mardak:
This means to handle resizing of link generator, the videoSize.realVideo\* state in MediaView should update on VideoDimensionsChanged and the videoSize.client\* state should change on `"resize"` events of the `<video>` element.

shared/css/common.css
```diff
@@ -591,3 +591,17 @@ html[dir="rtl"] .context-wrapper > .context-preview {
...
+
+.remote-video-box {
+  height: 100%;
+  width: 100%;
+}
+
+.remote-video-box > .remote-cursor {
+  background: url('../img/cursor.svg#cursor-orange') no-repeat;
+  height: 20px;
+  position: absolute;
```
Mardak, line 603: This was tricky to track down. Turns out the svg cursor has a white outline (similar to what OS X has around the black cursor). The colored point should be what we actually point at, so we need a `margin: -2px` here to have things match up better. Also include a comment to why we need the negative margin.

shared/js/remoteCursorStore.js
```diff
@@ -0,0 +1,106 @@
...
+          break;
+      }
+    },
+
+    /**
+     * Listen to stream dimension changes.
+     *
+     * @param {sharedActions.VideoDimensionsChanged} actionData
+     */
+    videoDimensionsChanged: function(actionData) {
```
Mardak, line 91: This is indeed the right action, but from my testing, this doesn't seem to trigger right away -- sometimes a minute after the screen video stream resized. @Standard8 @mikedeboer does this seem like an OTsdk issue? A quick glance seems to indicate this event is triggered when checking for time increments.. and perhaps there's no time data associated with screen sharing?

shared/js/views.jsx
```diff
@@ -831,6 +881,127 @@ loop.shared.views = (function(_, mozL10n) {
...
+        }
+      } else if (realVideoHeight >= clientHeight) {
+        streamVideoWidth = (realVideoWidth * clientHeight) / realVideoHeight;
+        streamVideoHeight = clientHeight;
+      } else {
+        // In this case we should keep the real dimension of the video since the
+        // video fits in the screen without reducing width or height
+        streamVideoWidth = realVideoWidth;
+        streamVideoHeight = realVideoHeight;
+      }
```
Mardak: This last else has the wrong assumption of what happens for small stream video sizes. It actually stretches to fit the available width or height instead of keeping the original size.

All of the conditionals can be simplified by comparing width/height ratios of the element and stream. I've committed this here: https://github.com/Mardak/loop/commit/23cbc16940b1ef0b1c038376525783128c2913cd and you can just integrate it into your commit.

Mardak: Here's the relevant new code:

```
      var clientRatio = clientWidth / clientHeight;
      var realVideoRatio = realVideoWidth / realVideoHeight;

      // If the video element ratio is "wider" than the video content, then the
      // full client height will be used and letterbox will be on the sides.
      // E.g., video element is 3:2 and stream is 1:1, so we end up with 2:2.
      var isWider = clientRatio > realVideoRatio;
      var streamVideoHeight = isWider ? clientHeight : clientWidth / realVideoRatio;
      var streamVideoWidth = isWider ? clientHeight * realVideoRatio : clientWidth;
```

shared/js/remoteCursorStore.js
Mardak, line 91: If we can't rely on timely videoDimensionsChanged, should we just pass the dimensions as part of the cursor event? (Although this doesn't fix the issue when the link generator previously moved the cursor then resizes the tab -- there wouldn't necessarily be a cursor event to have the cursor on link clicker side to update.)

shared/js/otSdkDriver.js
```diff
@@ -715,6 +719,45 @@ loop.OTSdkDriver = (function() {
...
+
+    /**
+     * Setups the subscriber cursor data channel that will handle the messages
+     * received with the cursor information about the real stream size and position
+     * of the remote cursor.
+     *
+     * @param {OT.Subscriber} sdkSubscriberObject The subscriber object for the stream.
+     */
+    _setupCursorDataChannel: function(sdkSubscriberObject) {
+      sdkSubscriberObject._.getDataChannel("cursor", {}, function(err, channel) {
```
Standard8, line 734: This and `_createCursorDataChannel` duplicate a lot of code.

Couldn't we do something like:

```
var dataChannels = ["cursor", "text"];
dataChannels.forEach(setupCursorDataChannel(channel))
```

That's only approximate, of course ;-)

shared/js/otSdkDriver.js
```diff
@@ -715,6 +719,45 @@ loop.OTSdkDriver = (function() {
...
+        // Sends will queue until the channel is fully open.
+        if (err) {
+          console.error(err);
+          this._notifyMetricsEvent("sdk.datachannel.sub." + err.message);
```
Standard8, line 738: I'd prefer us to have separate reports for each datachannel type, so maybe change this to `sdk.datachannel.<name>.sub`

shared/js/otSdkDriver.js
```diff
@@ -752,6 +795,35 @@ loop.OTSdkDriver = (function() {
...
+     * We create the cursor data channel once the remote client is setup for data
+     * channels. Getting the data channel for the subscriber is handled
+     * separately when the subscription completes.
+     */
+    _createCursorDataChannel: function() {
+      // This won't work until a subscriber exists for this publisher
+      this.publisher._.getDataChannel("cursor", {}, function(err, channel) {
+        if (err) {
+          console.error(err);
+          this._notifyMetricsEvent("sdk.datachannel.pub." + err.message);
```
Standard8, line 814: This should also change to `sdk.datachannel.<name>.pub`

shared/js/otSdkDriver.js
Mardak, line 734: Fixed. If only we could use fat arrow functions and destructured assignment to clean it up. ;)

shared/js/otSdkDriver.js
Mardak, line 738: Changed to include the type/name and updated tests to check for it.

Mardak:
I cleaned up the commits by moving various changes to the appropriate commit. I also cleaned up various comments (full sentences with punctuation). Also updated commit messages. And rebased on master.

Mardak:
Tested with production standalone, and it works as it does even when desktop is sending cursor positions.

Mardak:
Merged the updated commits in #131.

PR #91, opened 2016-01-19
Bug 1240554 - change mute display to match newer visual specs

chrafuse:
Option 1 - Sample of flex code volume circle and icon centring.


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

chrafuse:
Uses translate to centre within video port.  Sizes are set for particular media queries and video layouts through CSS.  See attached screen-shots in the bug of usage.

mozilla-autolander-deprecated:

chrafuse:
@dmose Corrected video size issues and using transform translate.  I will attach screen-shots on the bug.

mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

chrafuse:
@dmose - CSS for avatar sizes and video modes.  UI-review is good and attached screen-shots to the bug.

shared/css/conversation.css
```diff
@@ -364,11 +364,59 @@ html[dir="rtl"] .settings-menu.dropdown-menu {
...
  * For any audio-only streams, we want to display our own background
  */
+.avatar-container {
+  width:  100%;
+  height: 100%;
+  background-color: #d8d8d8;
+  overflow: hidden;
+}
+
+.media-wrapper > .remote > .avatar-container {
```
Standard8, line 374: General comment: the component you're dealing with here is the `AvatarView`, which in the code is set up as class `avatar-container`.

Therefore all css pertaining to the display of items within `avatar-container`, should only be referenced at the `avatar-container` level. The only things that should reference child selectors from other classes outside of it should be positional/layout items.

If, for example, you want to specify a different background colour for a remote element versus a local one, then you should expose a new class name on the `avatar-container`, e.g.

``` css
.avatar-container.remote {
  background-colour: #00a9dc
}
```

This way, the component can be re-used or moved to a different place in the hierarchy without needing to redo lots of css.

shared/css/conversation.css
```diff
@@ -364,11 +364,59 @@ html[dir="rtl"] .settings-menu.dropdown-menu {
...
+
+.media-wrapper .local .avatar-volume {
+  background-color: #e8e8e8;
+}
+
+.media-wrapper:not(.receiving-screen-share) > .remote .avatar-volume,
+.media-wrapper:not(.showing-remote-streams) > .focus-stream > .local
+{
+  height: 146px;
+  width: 146px;
```
Standard8, line 403: You've already specified these are the default size above, so I don't see you'd need to repeat them.

shared/css/conversation.css
```diff
@@ -416,6 +468,8 @@ html[dir="rtl"] .settings-menu.dropdown-menu {
 .remote > .avatar {
   /* make visually distinct from local avatar */
   opacity: 0.25;
+  background-image: url("../img/local-avatar-thumb.svg");
```
Standard8, line 471: This appears wrong, you're saying `.avatar` uses local, then `.remote .avatar` uses guest, then finally `.remote > .avatar` uses local.

If you change these three like I suggested above, then you'd end up with just two: `.avatar` with local, and `.avatar.remote` with guest.

shared/css/conversation.css
```diff
@@ -538,8 +592,14 @@ body[platform="win"] .share-service-dropdown.overflow > .dropdown-menu-item {
...
 
+.media-wrapper > .local .avatar-volume,
+.media-wrapper > .remote .avatar-volume{
+  width: 106px;
+  height: 106px;
+}
+
 .media-wrapper > .local .local-video,
-.media-wrapper > .focus-stream > .local .local-video {
+.media-wrapper > .remote.focus-stream > .local .local-video {
```
Standard8, line 602: This feels like a wrong change. The focus stream is not always the remote stream - sometimes its the screen sharing stream (especially on standalone). This feels like its breaking something.

shared/css/conversation.css
```diff
@@ -643,9 +709,39 @@ body[platform="win"] .share-service-dropdown.overflow > .dropdown-menu-item {
...
+    position: relative;
+    margin: 0;
+    bottom: auto;
+    width: 100%;
+    height: 100%;
+  }
+
+  .media-wrapper:not(.receiving-screen-share) > .focus-stream > .local > .avatar-volume {
+    height: 146px;
+    width: 146px;
```
Standard8, line 724: These are the same as the default size again, so shouldn't need specifying...

shared/css/conversation.css
```diff
@@ -643,9 +709,39 @@ body[platform="win"] .share-service-dropdown.overflow > .dropdown-menu-item {
...
+    border: 2px solid rgb(230, 230, 230);
+    border: 2px solid rgba(255, 255, 255, .9);
+    -webkit-background-clip: padding-box; /* for Safari */
+    background-clip: padding-box; /* for IE9+, Firefox 4+, Opera, Chrome */
     height: 120px;
   }
 
+  .media-wrapper.showing-remote-streams > .remote.focus-stream > .local .avatar-volume {
+    height: 106px;
+    width: 106px;
```
Standard8, line 737: I think it would be good to update the user story in the bug with the various situations & sizes the volume part should be. It'll be easier to review and ensure the css is correct that way.

ianb:
We're going to close this for now, until we have the new visuals

PR #95, opened 2016-01-21
Bug 1240512 - added onerror handler for POST requsts in standaloneMozLoop

Vidhuran:


Standard8:
nit:

`Bug 1240512 - added oneror handler...`

Spelling mistake in the commit message - should be `onerror`

standalone/test/standaloneMozLoop_test.js
```diff
@@ -181,6 +181,19 @@ describe("loop.StandaloneMozLoop", function() {
...
+
+    it("should call the callback on xhr error", function() {
+      var promise = loop.request("Rooms:Join", "fakeToken").then(function(result) {
+        expect(result.isError).eql(true);
+        expect(/HTTP 0/.test(result.message)).eql(true);
+      });
+
+      // Method to mock network failure
+      // https://github.com/sinonjs/sinon/issues/361
+      requests[0].respond(0, {}, "");
```
Standard8, line 191: I took a slightly more in-depth look at this, and realised that this wasn't actually testing the onerror handler (onload is still called). However, in the latest unpublished version of sinon, the issues have been fixed, and this will end up just calling onerror.

That's good news. I think we can leave this test code as-is, and we'll just pick up the new version of sinon when it is released.

Please could you make just a couple of small changes to drop the reference to sinon issue, as I don't think that's necessary any more. Just:

`// Method to mock a network failure.`

should be enough.

PR #98, opened 2016-01-21
Bug 1241649 - Hook up `npm version` to be able to bump the install.rdf.in version number and be ready to tag releases

Standard8:


Makefile
```diff
@@ -54,6 +54,26 @@ BABEL := $(NODE_LOCAL_BIN)/babel
...
+# - reduce to just the first of those lines
+# - strip out space, " and ,
+# - get the real version number (e.g. version:0.1.0 -> 0.1.0)
+# - change 0.1.0-alpha to 0.1.0alpha for AMO compatiblity
+PACKAGE_VERSION := $(shell cat package.json | \
+	grep version | \
+	head -1 | \
+	sed 's/[ ",]//g' | \
+	awk -F: '{ print $$2 }' | \
+	sed 's/-alpha/alpha/')
```
Mardak, line 66: Could combine the first `sed` and `awk` with `cut -d'"' -f4`

Makefile
```diff
@@ -54,6 +54,26 @@ BABEL := $(NODE_LOCAL_BIN)/babel
...
 
+# In the PACKAGE_VERSION below we:
+# - parse package.json
+# - get the lines with version in
+# - reduce to just the first of those lines
+# - strip out space, " and ,
+# - get the real version number (e.g. version:0.1.0 -> 0.1.0)
+# - change 0.1.0-alpha to 0.1.0alpha for AMO compatiblity
+PACKAGE_VERSION := $(shell cat package.json | \
+	grep version | \
```
Mardak: Any benefits of `cat file` then `grep` vs just `grep file`?

Makefile
```diff
@@ -54,6 +54,26 @@ BABEL := $(NODE_LOCAL_BIN)/babel
...
+	head -1 | \
+	sed 's/[ ",]//g' | \
+	awk -F: '{ print $$2 }' | \
+	sed 's/-alpha/alpha/')
+
+# Commands need to update the versions correctly in all places. Called from
+# npm's version command as configured in package.json
+.PHONY: update_version
+update_version:
+	sed -i '' -e 's/em:version>.*<\/em/em:version>$(PACKAGE_VERSION)<\/em/' $(ADD-ON)/install.rdf.in
```
Mardak: I'm for saving characters, but dropping the `<` of the opening tag and `>` of the closing makes it harder to parse. ;)

Makefile
```diff
@@ -54,6 +54,26 @@ BABEL := $(NODE_LOCAL_BIN)/babel
...
+# In the PACKAGE_VERSION below we:
+# - parse package.json
+# - get the lines with version in
+# - reduce to just the first of those lines
+# - strip out space, " and ,
+# - get the real version number (e.g. version:0.1.0 -> 0.1.0)
+# - change 0.1.0-alpha to 0.1.0alpha for AMO compatiblity
+PACKAGE_VERSION := $(shell cat package.json | \
+	grep version | \
+	head -1 | \
```
Mardak: `grep -m1` will limit to 1/first match

Standard8: Thank you for the optimisations!

Makefile
Standard8: I've made this multi-line as well, and added an `@` as it was unnecessary noise on the output.

PR #106, opened 2016-01-25
Makefile node modules directory regression fix

chrafuse:
@dmose Fixes node module directory regression.


mozilla-autolander-deprecated:

PR #108, opened 2016-01-25
Makefile fix using abspath

chrafuse:
@dmose using abspath to determine node_module location.


mozilla-autolander-deprecated:

dmose:
r=@dmose

PR #111, opened 2016-01-26
Bug 1242908 - Remove some obsolete configuration settings in the standalone server.

Standard8:


fcampo:
maybe we should get rid of `LOOP_FEEDBACK_API_URL` and `LOOP_FEEDBACK_PRODUCT_NAME` on the standalone `README.md` file too?

Standard8:
> maybe we should get rid of LOOP_FEEDBACK_API_URL and LOOP_FEEDBACK_PRODUCT_NAME on the standalone README.md file too?

We have extra READMEs? Argh!

Seriously, they're old and part of the previous documentation. I'll whip up a patch to remove all the old ones. Any documentation we need should be in the new structure.

PR #112, opened 2016-01-26
Bug 1154277 - Introduce a new pref for Loop e10s work

mikedeboer:
This PR contains all the changes necessary for e10s to work for the Hello add-on.

### DIFF
--- a/add-on/chrome/bootstrap.js
+++ b/add-on/chrome/bootstrap.js
@@ -95,16 +95,14 @@ var WindowListener = {
        *
        * @param {DOMEvent} [event] Optional event that triggered the call to this
        *                           function.
-       * @param {String}   [tabId] Optional name of the tab to select after the panel
-       *                           has opened. Does nothing when the panel is hidden.
        * @return {Promise}
        */
-      togglePanel: function(event, tabId = null) {
+      togglePanel: function(event) {
         if (!this.panel) {
           // We're on the hidden window! What fun!
           let obs = win => {
             Services.obs.removeObserver(obs, "browser-delayed-startup-finished");
-            win.LoopUI.togglePanel(event, tabId);
+            win.LoopUI.togglePanel(event);
           };
           Services.obs.addObserver(obs, "browser-delayed-startup-finished", false);
           return window.OpenBrowserWindow();
@@ -116,9 +114,10 @@ var WindowListener = {
           });
         }
 
-        return this.openCallPanel(event, tabId).then(doc => {
-          let fm = Services.focus;
-          fm.moveFocus(doc.defaultView, null, fm.MOVEFOCUS_FIRST, fm.FLAG_NOSCROLL);
+        return this.openPanel(event).then(mm => {
+          if (mm) {
+            mm.sendAsyncMessage("Social:EnsureFocusElement");
+          }
         }).catch(err => {
           Cu.reportError(err);
         });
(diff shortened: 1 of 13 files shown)


add-on/chrome/bootstrap.js
```diff
@@ -180,21 +145,17 @@ var WindowListener = {
               // have resumed the tour as soon as the visitor joined if it was (and
               // the pref would have been set to false already.
               this.MozLoopService.resumeTour("waiting");
-              resolve();
+              resolve(null);
```
Standard8, line 151: So interestingly, I've just realised that openCallPanel doesn't actually resolve the promise in the non-resume case. Not sure if that means we're leaving things hanging, but I think we should tidy that up here.

add-on/chrome/bootstrap.js
```diff
@@ -126,49 +125,15 @@ var WindowListener = {
        *
        * @param {event}  event   The event opening the panel, used to anchor
        *                         the panel to the button which triggers it.
-       * @param {String} [tabId] Identifier of the tab to select when the panel is
-       *                         opened. Example: 'rooms', 'contacts', etc.
        * @return {Promise}
        */
-      openCallPanel: function(event, tabId = null) {
+      openCallPanel: function(event) {
```
Standard8: Do you mind renaming this whilst you're touching it, e.g. openPanel or something? It feels better that way as its not just about calls...

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -796,13 +795,16 @@ var MozLoopServiceInternal = {
...
+   * @param {Object}        The peerConnection in question.
    */
-  stageForTelemetryUpload: function(window, pc) {
-    window.WebrtcGlobalInformation.getAllStats(allStats => {
-      let internalFormat = allStats.reports[0]; // filtered on pc.id
-      window.WebrtcGlobalInformation.getLogging("", logs => {
-        let report = convertToRTCStatsReport(internalFormat);
+  stageForTelemetryUpload: function(window, details) {
+    let mm = window.messageManager;
+    mm.addMessageListener("Loop:GetAllWebrtcStats", message => {
```
Standard8: Shouldn't the listener be removed somewhere?

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -996,72 +969,93 @@ var MozLoopServiceInternal = {
...
-          if (customSize && currSize != customSize && currSize != "loopChatMessageAppended") {
-            chatbox.setAttribute("customSize", customSize);
-            chatbox.parentNode.setAttribute("customSize", customSize);
-          }
-        }
-
-        window.addEventListener("LoopChatEnabled", onChatEvent);
-        window.addEventListener("LoopChatMessageAppended", onChatEvent);
-        window.addEventListener("LoopChatDisabledMessageAppended", onChatEvent);
+        mm.addMessageListener("Social:CustomEvent", message => {
```
Standard8: I can't find a remove for this message listener...

add-on/preferences/prefs.js
```diff
@@ -1,4 +1,5 @@
 pref("loop.enabled", true);
+pref("loop.remote.autostart", true);
```
Standard8: I'd suggest we land this as false initially, since it'll take a day or two to get the social patches into nightly & uplifted.

Standard8: and also until we get the crash sorted out.

add-on/chrome/modules/MozLoopService.jsm
mikedeboer: I believe listeners on the message manager share the objects' lifetime, thus are cleaned up when the messageManager is destroyed.
I'll ask around about that.

mikedeboer: Appears that we do need to remove it.

add-on/chrome/bootstrap.js
mikedeboer, line 151: We are, please check the `callback` function... or do you mean something else?

Standard8, line 151: Ah thanks, I'd missed that.

PR #116, opened 2016-01-26
Bug 1240554 ui showcase enable facebook fix

chrafuse:
@dmose @Standard8: Added tests for parseStringToElements and formatURL.  Only issue is that shared tests fails for the Unexpected Logged Warnings and Errors Check because parseStringToElements test that I added emits an error.  How do we send true for the suppressConsoleError in the LinkifiedTextView_test to not emit the error -OR- do we remove the test -OR- do we set parseStringToElements call to formatURL to suppressConsoleError(no errors ever emitted)?

### DIFF
--- a/shared/js/utils.js
+++ b/shared/js/utils.js
@@ -386,18 +386,18 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
     var urlObject;
     try {
       urlObject = new URL(url);
+      // Finally, ensure we look good.
+      return {
+        hostname: urlObject.hostname,
+        location: decodeURI(urlObject.href)
+      };
     } catch (ex) {
       if (suppressConsoleError ? !suppressConsoleError : true) {
-        console.error("Error occurred whilst parsing URL:", ex);
+        console.log("Error occurred whilst parsing URL: ", ex);
+        console.trace();
       }
       return null;
     }
-
-    // Finally, ensure we look good.
-    return {
-      hostname: urlObject.hostname,
-      location: decodeURI(urlObject.href)
-    };
   }
 
   /**
--- a/shared/test/utils_test.js
+++ b/shared/test/utils_test.js
@@ -296,7 +296,7 @@ describe("loop.shared.utils", function() {
   describe("#formatURL", function() {
     beforeEach(function() {
       // Stub to prevent console messages.
-      sandbox.stub(window.console, "error");
+      sandbox.stub(window.console, "log");
     });
 
     it("should decode encoded URIs", function() {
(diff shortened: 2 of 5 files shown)


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

chrafuse:
@dmose @Standard8: Changed to log and trace instead of error.

mozilla-autolander-deprecated:

chrafuse:
Reverted changes to fake data in UI showcase.  Renders the view.

shared/js/utils.js
```diff
@@ -386,18 +386,18 @@ var inChrome = typeof Components != "undefined" && "utils" in Components;
...
+      // Finally, ensure we look good.
+      return {
+        hostname: urlObject.hostname,
+        location: decodeURI(urlObject.href)
+      };
     } catch (ex) {
       if (suppressConsoleError ? !suppressConsoleError : true) {
-        console.error("Error occurred whilst parsing URL:", ex);
+        console.log("Error occurred whilst parsing URL: ", ex);
+        console.trace();
```
Standard8, line 397: I think we can leave this as console.error - that gives a trace automatically.

shared/test/utils_test.js
```diff
@@ -318,14 +318,17 @@ describe("loop.shared.utils", function() {
         });
     });
 
-    it("should return null if it the url is not valid", function() {
-      expect(sharedUtils.formatURL("hinvalid//url")).eql(null);
+    it("should return null if suppressConsoleError is true and the url is not valid", function() {
+      expect(sharedUtils.formatURL("hinvalid//url", true)).eql(null);
```
Standard8, line 322: No need to pass `true` here. The stub in the before each should prevent issues.

shared/test/utils_test.js
```diff
@@ -318,14 +318,17 @@ describe("loop.shared.utils", function() {
...
     });
 
-    it("should log an error message to the console", function() {
-      sharedUtils.formatURL("hinvalid//url");
+    it("should return null if suppressConsoleError is true and is a malformed URI sequence", function() {
+      expect(sharedUtils.formatURL("http://www.example.com/DNA/pizza/menu/lots-of-different-kinds-of-pizza/%8D%E0%B8%88%E0%B8%A1%E0%B8%A3%E0%8D%E0%B8%88%E0%B8%A1%E0%B8%A3%E0%", true)).eql(null);
```
Standard8, line 326: No need to pass `true` here. The stub in the before each should prevent issues.

Standard8:
I'd suggest making the commit message a little clearer, e.g. `Bug 1240554 - Fix layout bustage in the ui-showcase by providing defaults for the standalone button. Also fix errors from an invalid url.`

shared/js/utils.js
Standard8, line 397: Ok, per dmose on irc, leave it as `console.log`, but lets drop the `console.trace`.

Standard8:
Oh, also, since this isn't actually bug 1240554, I'd just drop the bug number.

r=Standard8.

Don't forget to force-push your rebased branch to here before you --ff-only merge it to master.

mozilla-autolander-deprecated:

PR #119, opened 2016-01-27
Remove obsolete README files.

Standard8:
These are all files that are now obsolete, and where necessary, the content is already in Developing.md. So I think we should just remove these to save confusion.


mozilla-autolander-deprecated:

Standard8:
rs=@dmose over irc.

mozilla-autolander-deprecated:

PR #126, opened 2016-01-29
Bug 1244119 - Improve the release process; handle alpha version better and improve locale imports.

Standard8:


Makefile
```diff
@@ -11,7 +11,7 @@ LOOP_DOWNLOAD_FIREFOX_URL := $(shell echo $${LOOP_DOWNLOAD_FIREFOX_URL-"https://
 LOOP_PRIVACY_WEBSITE_URL := $(shell echo $${LOOP_PRIVACY_WEBSITE_URL-"https://www.mozilla.org/privacy/firefox-hello/"})
 LOOP_LEGAL_WEBSITE_URL := $(shell echo $${LOOP_LEGAL_WEBSITE_URL-"https://www.mozilla.org/about/legal/terms/firefox-hello/"})
 LOOP_PRODUCT_HOMEPAGE_URL := $(shell echo $${LOOP_PRODUCT_HOMEPAGE_URL-"https://www.firefox.com/hello/"})
-FIREFOX_VERSION=45.0
+FIREFOX_VERSION=47.0
```
Mardak, line 14: Does Firefox even check maxVersion? Or is this more for AMO listing showing the correct version -- but even then, I thought AMO had automatic maxVersion bumping. I suppose figuring out a way to not need to do this every 6 weeks is a separate issue.

Makefile
```diff
@@ -64,19 +64,33 @@ PACKAGE_VERSION := $(shell grep -m1 version package.json | \
 	cut -d'"' -f4 | \
 	sed 's/-alpha/alpha/')
 
-# Commands need to update the versions correctly in all places. Called from
-# npm's version command as configured in package.json
 .PHONY: update_version
-update_version: $(VENV)
+generate_changelog: $(VENV)
```
Mardak, line 68: I believe this `.PHONY` should be for `generate_changelog`?

Makefile
```diff
@@ -64,19 +64,33 @@ PACKAGE_VERSION := $(shell grep -m1 version package.json | \
...
+.PHONY: change_versions
+change_versions:
 	@# Ubuntu's version of sed doesn't have -i
 	@sed -e 's/<em:version>.*<\/em:version>/<em:version>$(PACKAGE_VERSION)<\/em:version>/' \
 	    $(ADD-ON)/install.rdf.in > $(ADD-ON)/install.rdf.in.gen
 	@mv $(ADD-ON)/install.rdf.in.gen $(ADD-ON)/install.rdf.in
-	@$(VENV)/bin/gitchangelog | sed -e 's/%%version%% (unreleased)/${PACKAGE_VERSION}/' > CHANGELOG.md
-	@git add CHANGELOG.md $(ADD-ON)/install.rdf.in
+	# Add package.json for the case where `npm version` is used with
+	# `--no-git-tag-version` - to make it easier to commit.
```
Mardak: These have extra leading tabs if you wanted it to be a comment.

bin/locale_update.py
```diff
@@ -35,7 +41,15 @@ def create_locale(src_dir):
...
-        shutil.copytree(os.path.join(l10n_src, src_dir), os.path.join(l10n_dst, dst_dir))
+        # Don't copy the en-US files. Stick with what's in our tree as that might
+        # be a later version.
+        if dst_dir == "en-US":
+            return dst_dir
+
+        # Copy the l10n files, but ignore any `.keep`
+        shutil.copytree(os.path.join(l10n_src, src_dir),
+                        os.path.join(l10n_dst, dst_dir),
+                        ignore=shutil.ignore_patterns(".keep"))
```
Mardak: Nice. I suppose while we're at it, should we update Makefile to have a target that runs the script and `@git add locale` ?

Standard8: I haven't quite decided yet. I'm sort-of thinking that we need to do visual checks of the diffs. Though at the same time, I'm pondering automation for that...

Makefile
Standard8, line 14: Yeah, I'm not quite sure what the status of all this is now. I've got a few ideas in mind for how to make it easier on ourselves.

AMO does have automatic version bumping, but that doesn't help when you first upload it.

bin/locale_update.py
```diff
@@ -35,7 +41,15 @@ def create_locale(src_dir):
         # Convert loop-client-l10n repo names to loop repo names.
         dst_dir = src_dir.replace('_', '-').replace('templates', 'en-US')
 
-        shutil.copytree(os.path.join(l10n_src, src_dir), os.path.join(l10n_dst, dst_dir))
+        # Don't copy the en-US files. Stick with what's in our tree as that might
+        # be a later version.
+        if dst_dir == "en-US":
+            return dst_dir
```
Mardak: Probably flip the condition and move copytree inside. Both paths in this commit `return dst_dir`.

Makefile
Mardak: Ah yeah you can use `@#` like the `@# Ubuntu's...` above.

bin/locale_update.py
Mardak: You could always do `git diff --cached` to check the diff as one would need to do with the other automated `git add`s.

PR #130, opened 2016-01-30
Hack our way around functional test breakage caused by Ubuntu auto-in…

dmose:
…stalling add-ons.

Going to let @Standard8 have a look at this.  I'm pretty surprised that turning off the blocklist and the xpinstall signature stuff was necessary for the Ubuntu add-on.  Or was that to fix the fact that we're now a system add-on?  Question is really for @nils-ohlmeier...

### DIFF
--- a/test/functional/config.py
+++ b/test/functional/config.py
@@ -19,5 +19,14 @@ FIREFOX_PREFERENCES = {
     # this dialog is fragile, and likely to introduce intermittent failures
     "media.navigator.permission.disabled": True,
     # Use fake streams only
-    "media.navigator.streams.fake": True
+    "media.navigator.streams.fake": True,
+
+    # attempts to work around Ubuntu wanting to install add-ons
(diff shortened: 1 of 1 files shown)


mozilla-autolander-deprecated:

nils-ohlmeier:
Sorry I forgot to mention in my email that after a few initial attempts with only turning off the xpinstall signature stuff that I got sick of spending more time on this and pulled out the big gun: I simply copied all the extension settings which looked vaguely relevant from testing/profiles/prefs_general.js.
So it is quite possible that a smaller subset of these settings would accomplish the same.

Standard8:
I'm actually fine with all of these prefs - I think they are worth having especially things like disabling extension update checks & blocklist items - i.e. things that could happen in the background and mess up the test.

r=Standard8

mozilla-autolander-deprecated:

dmose:
Merged.

PR #133, opened 2016-02-01
Fix the tiles support url used for local standalone server instances

Standard8:
This fixes the URL used for our local instances of standalone to be the current one for the support about the tiles. Hence, matching to production and saving any confusion.


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

PR #134, opened 2016-02-01
Bug 1244539 - Pointer still gets shared when tab sharing is stopped/p…

mancas:
…aused. r=Mardak

### DIFF
--- a/add-on/chrome/bootstrap.js
+++ b/add-on/chrome/bootstrap.js
@@ -660,6 +660,11 @@ var WindowListener = {
        * through the sdk.
        */
       handleMousemove: function(event) {
+        // We want to stop sending events if sharing is paused.
+        if (this._browserSharePaused) {
+          return;
+        }
+
         // Only update every so often.
         let now = Date.now();
         if (now - this.lastCursorTime < MIN_CURSOR_INTERVAL) {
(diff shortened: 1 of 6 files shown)


Standard8:
Just happened to glance by this. I'd expect us to stop sending the cursor position/events when sharing is paused as well. Whilst I can't think of any privacy implications, I think I'd rather we stop all screen share events over the network if sharing is paused.

Mardak:
I agree that we shouldn't send the event in the first place. This actually should simplify the code quite a bit. `bootstrap.js` already has `this._browserSharePaused`, so `handleMousemove` should be able to just short circuit.

Standard8:
We'll still want to remove the pointer from the display... and we are going to need some alternate display on the link clicker when the sharing is paused later on, so we definitely want what's here as well.

shared/js/remoteCursorStore.js
```diff
@@ -44,7 +45,8 @@ loop.store.RemoteCursorStore = (function() {
     getInitialStoreState: function() {
       return {
         realVideoSize: null,
-        remoteCursorPosition: null
+        remoteCursorPosition: null,
+        shouldRenderCursor: true
```
Mardak: At first I was going to suggest just replacing remoteCursorPosition when hasVideo goes false, but keeping the position separate would allow it to persist its location to show. Although it's not necessarily useful to have the cursor reshow where it was before the sharing was paused, but that's a minor issue.

add-on/chrome/bootstrap.js
```diff
@@ -660,6 +660,11 @@ var WindowListener = {
        * through the sdk.
        */
       handleMousemove: function(event) {
+        // We want to stop sending events if sharing is paused
```
Mardak: Comment with full sentences, so end with a period (and remove trailing space).

PR #136, opened 2016-02-02
Bug 1244744 - Pointer is not visible in standalone on Chrome

Mardak:
r?@mancas 

The main style change commented here because we compress the svg:

```
use, /* Hide all use by default */
:target ~ #orange /* Hide the default (last) #orange if there's a :target */ {
  display: none;
}

#orange, /* Have #orange be shown by default */
:target /* Show the :target if there's one */ {
  display: inherit;
}
```

Also renamed #cursor-orange to just #orange as cursor.svg#orange should be clear enough and more compact. Moved #orange to be the last because it needs to be a ~ sibling to be hidden.


mancas:
Looks good! r=me but before landing we need to fix travis. It's failing when trying to install google-chrome. Please take a look at the error.

Thank you

PR #140, opened 2016-02-02
Bug 1239828 - Make export work for L10n

Standard8:


add-on/jar.mn
```diff
@@ -118,7 +53,16 @@
 % override chrome://loop/skin/toolbar@2x.png      chrome://loop/skin/toolbar-aero@2x.png          os=WINNT osversion=6.1
 % override chrome://loop/skin/toolbar@2x.png      chrome://loop/skin/toolbar-win8@2x.png         os=WINNT osversion=6.2
 % override chrome://loop/skin/toolbar@2x.png      chrome://loop/skin/toolbar-win8@2x.png         os=WINNT osversion=6.3
-  locale/                        (chrome/locale/*)
+  # Due to the way Loop's L10n works, we include a fallback to en-US.
+  content/locale/en-US/          (chrome/locale/en-US/*)
```
Mardak: To be clear, this /needs/ to be packaged outside of the locale directory because there's checks elsewhere to make sure there's only one packaged locale? Does it make sense to be explicit in the packaging by putting at `content/locale-fallback/`? Otherwise other places that reference `content/locale/en-US` might seem like a misplaced locale.

bin/generateChromeManifest.py
```diff
@@ -0,0 +1,101 @@
...
+
+    locale_dirs = sorted(os.listdir(l10n_src))
+
+    input_file = io.open(input_file_name, "r")
+
+    jar_mn = input_file.read()
+
+    # sed -n -e '/^%/p' $(ADD-ON)/jar.mn | \
+    # sed -e "s/^% //g" | sed -e "s/%/chrome\//g" \
+
```
Mardak: These `sed` comments probably aren't as useful as the expanded comments below. ;) So just remove them.

bin/generateChromeManifest.py
```diff
@@ -0,0 +1,101 @@
...
+    # Drop any line that doesn't begin with %.
+    manifest_output = "\n".join(line for line in jar_mn.split("\n") if (re.match("^%", line)))
+
+    # Drop the "% " from the start of lines.
+    manifest_output = re.sub(
+        '^% (.*$)',
```
Mardak: Some reason this seems odd with the $ inside ()s.. but I guess it's fine.

bin/generateChromeManifest.py
```diff
@@ -0,0 +1,101 @@
...
+        '(locale loop .*\n)',
+        '\n'.join(localeLines) + "\n",
+        manifest_output)
+
+    # For the loop-locale-fallback we need to adjust the path for inclusion.
+    # Mozilla-central actually duplicates file, the add-on by itself doesn't.
+    manifest_output = re.sub(
+        '(chrome/content/locale/en-US/)',
+        'chrome/locale/en-US/',
+        manifest_output)
```
Mardak, line 57: Do we just want to duplicate `loop.properties` file to `chrome/content/locale/en-US` (or `chrome/content/locale-fallback`) so the packaged add-on is more like the m-c add-on? Although m-c add-on will only have one locale, so keeping this similar might not be worth it.

bin/generateChromeManifest.py
```diff
@@ -0,0 +1,101 @@
...
+
+    print (manifest_output)
```
Mardak: Was this for debugging? Or do we want it to show up when building chrome.manifest?

bin/generateChromeManifest.py
```diff
@@ -0,0 +1,101 @@
...
+    #            jar_mn)
+
+    # One big if statement to avoid lots of if/endif lines, and the preprocessor
+    # can't cope with '\' on the end of the line.
+    # jar_locales = ['AB_CD == {0}'.format(x.replace("-", "_")) for x in locale_list]
+
+    # new_content = re.sub(
+    #     '(#if AB_CD.+\n)',
+    #     '#if ' + ' || '.join(jar_locales) + '\n',
+    #     new_content)
```
Mardak: This was moved to locale_update.py

bin/locale_update.py
```diff
@@ -72,12 +72,27 @@ def create_locale(src_dir):
...
+        # in so it can handle '_' instead.
+        dashLocales = []
+        for locale in locale_list:
+            if "-" in locale:
+                dashLocales.append('#define {0} {1}'.format(locale.replace("-", "_"), locale))
+
         new_content = re.sub(
-            '(% locale loop .+\n)+',
-            '\n'.join(jar_locales) + '\n',
+            '(#define .+\n)+',
```
Mardak, line 83: This could be dangerous if we needed to `#define` other stuff, but `#define LOCALE_zh_TW zh-TW` seems a bit silly. (Where before `% locale loop` had a bit more "context" that this was something specific to locales.) This should be fine for now.

bin/generateChromeManifest.py
```diff
@@ -0,0 +1,101 @@
...
+    manifest_output = re.sub(
+        '%',
+        'chrome/',
+        manifest_output, 0)
+
+    localeLines = ['locale loop {0} chrome/locale/{0}/'.format(x) for x in locale_dirs]
+
+    manifest_output = re.sub(
+        '(locale loop .*\n)',
+        '\n'.join(localeLines) + "\n",
```
Mardak: nit: here and elsewhere, there's a mix of `'\n'` (single quotes) and `"\n"` (double quotes).

add-on/jar.mn
Standard8: Yes that's right. I think you're right, its reasonable to move to locale-fallback. Probably will keep the en-US though just for clarity.

bin/generateChromeManifest.py
Standard8: Yeah, they were the old ones for reference, forgot to take them out.

bin/generateChromeManifest.py
Standard8: Yep, more debugging I forgot to remove...

bin/generateChromeManifest.py
Standard8, line 57: Yeah, I think these are different enough that we don't duplicate for the packaged add-on.

PR #144, opened 2016-02-03
Bug 1239970 - Display the FTU picture and the copy in the empty roomlist

daveccrit:
Requesting review from one of the following, whoever is available:
@dmose 
@Standard8 
@Mardak 

I also made a change to the UI Showcase to allow time for the iframes to complete changing height before scrolling to the hashtag.

### DIFF
--- a/locale/en-US/add-on.properties
+++ b/locale/en-US/add-on.properties
@@ -28,13 +28,15 @@ sign_in_again_use_as_guest_button2=Use {{clientSuperShortname}} as a Guest
 panel_browse_with_friend_button=Browse this page with a friend
 panel_disconnect_button=Disconnect
 
-## LOCALIZATION_NOTE(first_time_experience_subheading2): Message inviting the
+## LOCALIZATION_NOTE(first_time_experience_subheading2, first_time_experience_subheading_button_above): Message inviting the
 ## user to create his or her first conversation.
 first_time_experience_subheading2=Click the Hello button to browse Web pages with a friend.
+first_time_experience_subheading_button_above=Click on the button above to browse Web pages with a friend.
 
-## LOCALIZATION_NOTE(first_time_experience_content): Message describing
+## LOCALIZATION_NOTE(first_time_experience_content, first_time_experience_content2): Message describing
 ## ways to use Hello project.
 first_time_experience_content=Use it to plan together, work together, laugh together.
+first_time_experience_content2=Use it to get things done: plan together, laugh together, work together.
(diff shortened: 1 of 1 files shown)


locale/en-US/add-on.properties
```diff
@@ -28,13 +28,15 @@ sign_in_again_use_as_guest_button2=Use {{clientSuperShortname}} as a Guest
 panel_browse_with_friend_button=Browse this page with a friend
 panel_disconnect_button=Disconnect
 
-## LOCALIZATION_NOTE(first_time_experience_subheading2): Message inviting the
+## LOCALIZATION_NOTE(first_time_experience_subheading2, first_time_experience_subheading_button_above): Message inviting the
 ## user to create his or her first conversation.
 first_time_experience_subheading2=Click the Hello button to browse Web pages with a friend.
+first_time_experience_subheading_button_above=Click the button above to browse web pages with a friend.
```
dmose: This string is unnecessary.  Take a look at the conversation in the bug between Matej & RT.

dmose:
r=dmose with the first string removed as per bug discussion.

locale/en-US/add-on.properties
dmose: OK, re-reading the bug, it's possible that this needs to stay, though I don't think so.  So let's keep it to be safe.

However, s/Click the button/Click on the button/ and change web to Web as per the text in the bug.

locale/en-US/add-on.properties
```diff
@@ -28,13 +28,15 @@ sign_in_again_use_as_guest_button2=Use {{clientSuperShortname}} as a Guest
 panel_browse_with_friend_button=Browse this page with a friend
 panel_disconnect_button=Disconnect
 
-## LOCALIZATION_NOTE(first_time_experience_subheading2): Message inviting the
+## LOCALIZATION_NOTE(first_time_experience_subheading2, first_time_experience_subheading_button_above): Message inviting the
 ## user to create his or her first conversation.
 first_time_experience_subheading2=Click the Hello button to browse Web pages with a friend.
+first_time_experience_subheading_button_above=Click on the button above to browse Web pages with a friend.
```
dmose, line 34: Shouldn't this be first_time_experience_subheading3 ?

dmose:
r=dmose with one more change... :-)

locale/en-US/add-on.properties
dmose, line 34: Never mind; I see what's going on now.  r=dmose

PR #156, opened 2016-02-08
Bug 1246621 - Fix functional tests following e10s changes.

Standard8:


test/functional/serversetup.py
```diff
@@ -62,8 +65,8 @@ def start_loop_server(loop_server_location):
     def start_content_server():
         content_server_location = os.environ.get('STANDALONE_SERVER')
         if content_server_location is None:
-          content_server_location = os.path.join(os.path.dirname(__file__),
-                                                 "../../standalone")
+            content_server_location = os.path.join(os.path.dirname(__file__),
```
mikedeboer, line 68: These are linting changes, I presume?

test/functional/test_1_browser_call.py
```diff
@@ -81,7 +81,7 @@ def switch_to_chatbox(self):
         # for Marionette bug 1094246 to be fixed.
         chatbox = self.wait_for_element_exists(By.TAG_NAME, 'chatbox')
         script = ("return document.getAnonymousElementByAttribute("
-                  "arguments[0], 'class', 'chat-frame');")
+                  "arguments[0], 'anonid', 'content');")
```
mikedeboer, line 84: ...and this would be the actual fix, no?

test/functional/serversetup.py
Standard8, line 68: Yes, just fixing the indentation.

test/functional/test_1_browser_call.py
Standard8, line 84: Half of it, there's a similar change a bit further down. Its also why it is in a separate commit...

PR #157, opened 2016-02-08
Bug 1229195 - MozLoopService#openChatWindow should return a Promise t…

mikedeboer:
…hat is resolved with the window ID. r=Standard8


mikedeboer:
Merged with 991273be3ffcdbd111959a0a2e072a36fdd67d25

PR #160, opened 2016-02-09
Bug 1240516 - Add in-chat notification on peer disconnect by handling RemotePeerDisconnect event in textChatStore

Vidhuran:


Vidhuran:
I'll get a ui review before updating the test and ui showcase

shared/css/conversation.css
```diff
@@ -967,6 +967,31 @@ html[dir="rtl"] .text-chat-entry.received > p:after {
   align-self: auto;
 }
 
+.text-chat-notif {
+  display: flex;
+}
+
+.text-chat-notif > .content-wrapper {
+  margin-left: 10px;
```
Standard8, line 1039: Note: where you've got margin-left and margin-right (below), you'll need to provide [html="rtl"] versions of these to flip the borders for rtl languages. You can test this in the ui-showcase (there's an rtl checkbox at the top of the page)

shared/css/conversation.css
```diff
@@ -967,6 +967,31 @@ html[dir="rtl"] .text-chat-entry.received > p:after {
...
+  margin-bottom: 10px;
+  display: flex;
+  border: 1px solid #B1B1B1;
+  max-width: 70%;
+}
+
+.text-chat-notif img {
```
Standard8: Please use child selectors for these, they have better performance even if you have to be a bit more explicit.

shared/js/textChatView.jsx
```diff
@@ -79,6 +80,23 @@ loop.shared.views.chat = (function(mozL10n) {
         );
       }
 
+      if (this.props.contentType === CHAT_CONTENT_TYPES.NOTIFICATION) {
+
+        var thumbnail = this.props.useDesktopPaths ?
+            "shared/img/leave_notification.svg" :
+            "shared/img/leave_notification.svg";
```
Standard8: No need for this since these are the same now. In fact, I've just raised a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1247190) to drop the use of useDesktopPaths.

shared/css/conversation.css
```diff
@@ -1030,6 +1030,47 @@ html[dir="rtl"] .text-chat-entry.received > p:after {
...
+.content-wrapper > p {
+  margin-top: 8px;
+  margin-right: 10px;
+  margin-bottom: 8px;
+  color: #757575;
+  font-size: 12px;
+}
+
+html[dir="rtl"] .text-chat-notif > .content-wrapper {
+  margin-right: 10px;
```
Standard8, line 1057: Please add a line for `margin-left: 0px` - you need this to cancel out the margin-left that the original rule added.

shared/css/conversation.css
```diff
@@ -1030,6 +1030,47 @@ html[dir="rtl"] .text-chat-entry.received > p:after {
...
+}
+
+html[dir="rtl"] .content-wrapper > .notification-icon {
+  margin-top: 8px;
+  margin-right: 10px;
+  margin-left: 5px;
+  margin-bottom: 8px;
```
Standard8: Why not do this ruleset in the `margin: 8px 10px 8px 5px` similar to the original rule?

shared/css/conversation.css
```diff
@@ -1030,6 +1030,47 @@ html[dir="rtl"] .text-chat-entry.received > p:after {
...
+}
+
+html[dir="rtl"] .content-wrapper > p {
+  margin-left: 10px;
+  margin-right: 0px;
+}
+
+
+
```
Standard8: Nit: Only one blank line, or maybe 2 necessary here.

shared/js/textChatView.jsx
```diff
@@ -79,6 +80,21 @@ loop.shared.views.chat = (function(mozL10n) {
         );
       }
 
+      if (this.props.contentType === CHAT_CONTENT_TYPES.NOTIFICATION) {
+
+        var thumbnail = "shared/img/leave_notification.svg";
```
Standard8: nit: I think we can drop the intermediate variable now this has been simplified.

PR #161, opened 2016-02-09
Bug 1239972 - Adapt infobar message link clicker is in room or not

chrafuse:
@dmose Strings commit for infobar changes and are only for link generator.


dmose:
r=dmose

PR #163, opened 2016-02-09
Bug 1242706 - Make sharing affordances in desktop invitation view clearer and modal. String changes

daveccrit:
String changes. Review: @dmose


dmose:
My mistake; I see we've already got that string.  r=dmose

PR #165, opened 2016-02-10
Bug 1220627 - As a desktop client user, I want to be able to share si…

mancas:
…mply the Hello URL through a sharing panel after I create a room.


add-on/panels/js/panel.jsx
```diff
@@ -894,6 +885,114 @@ loop.panel = (function(_, mozL10n) {
...
+          <div className="share-panel-content">
+            <h2>{mozL10n.get("share_panel_heading")}</h2>
+            <p>
+              {mozL10n.get("share_panel_subheading")}
+            </p>
+
+            <div className="share-panel-link">
+              <span>{mozL10n.get("share_panel_link")}</span>
+              <input type="text" value={activeRoom.roomUrl} />
+              <sharedViews.CopyLinkButton
```
Standard8: I think its good that you've shared the buttons, but I think we should go a bit further. I think we should share probably all of `share-panel-content`.

The layout should be identical - it might be worth checking with UX if they really do want different text sizes etc between the two or not. If we do need to have slightly different sized text / icons then we should do that as properties (e.g. textSize="small") on the react element, translating to css classes that are then styled appropriately for the component.

I know the link text box is different, but I don't see a reason for it to be different between them - we should keep the same structure.

shared/js/views.jsx
```diff
@@ -508,6 +509,144 @@ loop.shared.views = (function(_, mozL10n) {
     }
   });
 
+  var CopyLinkButton = React.createClass({
```
Standard8: As we're sharing these only across desktop, and not standalone, I'd prefer for them to go into a `desktopViews.jsx` or something under the add-on directory.

Otherwise, we're adding more styles & code to the standalone which we don't need (and potentially confuses people as to where it is used).

Standard8:
We'll also want a view in the ui-showcase covering the share panel view.

add-on/panels/css/panel.css
```diff
@@ -775,3 +784,37 @@ img.fte-hello-web-share {
...
+  background-color: rgba(0, 0, 0, .3);
+  bottom: 0;
+  display: none;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+}
+
+.share-panel-container > .room-invitation-overlay {
```
Standard8: Testing it out, I'm not seeing the css being applied in the panel. I think it needs moving from conversation.css to probably a new css file that's shared just for desktop.

ui/ui-showcase.jsx
```diff
@@ -754,6 +755,19 @@
 
             <FramedExample cssClass="fx-embedded-panel"
                            dashed={true}
+                           height={250}
+                           summary="Share panel"
+                           width={330}>
+              <div className="panel">
+                <SharePanelView dispatcher={dispatcher}
```
Standard8, line 775: For some reason, this just gives a grey square for me when I view it.

locale/en-US/add-on.properties
```diff
@@ -191,6 +191,11 @@ rooms_list_currently_browsing2=Currently browsing
 rooms_signout_alert=Open conversations will be closed
 room_name_untitled_page=Untitled Page
 
+# Share panel
+share_panel_heading=Invite a friend to join you!
+share_panel_subheading=It takes two people to use Firefox Hello, so send a friend a link to browse the web with you!
+share_panel_link=Your link:
```
Standard8: I believe these additions are now unused.

add-on/panels/js/desktopViews.jsx
```diff
@@ -0,0 +1,304 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var loop = loop || {};
+loop.shared = loop.shared || {};
+loop.shared.desktopViews = (function(_, mozL10n) {
```
Standard8: I think `_` doesn't get used in this file, so can you drop it from here and the end please.

add-on/panels/js/desktopViews.jsx
```diff
@@ -0,0 +1,304 @@
...
+    statics: {
+      TRIGGERED_RESET_DELAY: 2000
+    },
+
+    mixins: [React.addons.PureRenderMixin],
+
+    propTypes: {
+      callback: React.PropTypes.func,
+      dispatcher: React.PropTypes.instanceOf(loop.Dispatcher).isRequired,
+      from: React.PropTypes.string.isRequired,
```
Standard8: `from` is a bit vague here. I know its vague in the action as well, but maybe we could at least add a bit of documentation or change the prop name. How about `buttonLocation` or `locationForMetrics` or something along one of those lines?

add-on/panels/js/desktopViews.jsx
```diff
@@ -0,0 +1,304 @@
...
+    }
+  });
+
+  var EmailLinkButton = React.createClass({
+    mixins: [React.addons.PureRenderMixin],
+
+    propTypes: {
+      callback: React.PropTypes.func,
+      dispatcher: React.PropTypes.instanceOf(loop.Dispatcher).isRequired,
+      from: React.PropTypes.string.isRequired,
```
Standard8: ditto with this `from`

add-on/panels/js/panel.jsx
```diff
@@ -891,6 +883,103 @@ loop.panel = (function(_, mozL10n) {
...
+        roomToken: activeRoom.roomToken
+      }));
+
+      this.closeWindow();
+    },
+
+    render: function() {
+      var sharePanelClasses = classNames({
+        "share-panel-container": true,
+        "share-panel-open": this.state.shouldOpenSharePanel || this.props.forceRender
```
Standard8: I would prefer if we didn't render the share panel at all if we're not displaying it.

add-on/panels/js/panel.jsx
```diff
@@ -891,6 +883,103 @@ loop.panel = (function(_, mozL10n) {
...
+      // the room.
+      if (this.state.pendingCreation &&
+          !nextState.pendingCreation && !nextState.error) {
+        this.setState({
+          shouldOpenSharePanel: true
+        });
+      }
+    },
+
+    _closeSharePanel: function() {
```
Standard8: How about naming this `handleClosing`? - it doesn't close the panel, and the `SharePanel` part is a duplicate of the component name.

add-on/panels/js/panel.jsx
```diff
@@ -891,6 +883,103 @@ loop.panel = (function(_, mozL10n) {
...
+      this.state.shouldOpenSharePanel && this._closeSharePanel();
+    },
+
+    componentWillUpdate: function(nextProps, nextState) {
+      // If we've just created a room, close the panel - the store will open
+      // the room.
+      if (this.state.pendingCreation &&
+          !nextState.pendingCreation && !nextState.error) {
+        this.setState({
+          shouldOpenSharePanel: true
```
Standard8: How about `showPanel`?  As it is really used to key if we're showing the panel or not.

add-on/panels/js/panel.jsx
Standard8: Actually, I take that back, `_closeSharePanel` is reasonable appropraite. I think I'd go more for `_closePanel` or `handleClosePanel` and also move the `this.closeWindow()` into this function - it wasn't obvious that this function was going to do that as well.

add-on/panels/js/desktopViews.jsx
```diff
@@ -0,0 +1,304 @@
...
+          </div>
+          <div className="input-button-group">
+            <div className="input-button-group-label">{mozL10n.get("invite_your_link")}</div>
+            <div className="input-button-content">
+              <div className="input-group group-item-top">
+                <input readOnly={true} type="text" value={this.props.roomData.roomUrl} />
+              </div>
+              <CopyLinkButton
+                dispatcher={this.props.dispatcher}
+                from="conversation"
```
Standard8: This needs to change to `{this.props.from}`

add-on/panels/js/desktopViews.jsx
```diff
@@ -0,0 +1,304 @@
...
+          })}>
+            <EmailLinkButton
+              dispatcher={this.props.dispatcher}
+              from="conversation"
+              roomData={this.props.roomData} />
+            {(() => {
+              if (this.props.facebookEnabled) {
+                return (<FacebookShareButton
+                          dispatcher={this.props.dispatcher}
+                          from="conversation"
```
Standard8: This needs to change to `{this.props.from}`

add-on/panels/js/desktopViews.jsx
```diff
@@ -0,0 +1,304 @@
...
+      error: React.PropTypes.object,
+      facebookEnabled: React.PropTypes.bool.isRequired,
+      from: React.PropTypes.string.isRequired,
+      // This data is supplied by the activeRoomStore.
+      roomData: React.PropTypes.object.isRequired,
+      show: React.PropTypes.bool.isRequired,
+      socialShareProviders: React.PropTypes.array
+    },
+
+    handleCopyButtonClick: function(event) {
```
Standard8: This function is now redundant since it moved into CopyLinkButton

add-on/panels/test/desktopViews_test.js
```diff
@@ -0,0 +1,352 @@
...
+        callback: callback
+      });
+      var facebookBtn = view.getDOMNode();
+      React.addons.TestUtils.Simulate.click(facebookBtn);
+
+      sinon.assert.calledOnce(callback);
+    });
+  });
+
+  describe("SharePanelView", function() {
```
Standard8, line 220: Please can you also add a couple of tests to check that it isn't displayed when the prop `show` is false, nor when `this.props.roomData.roomUrl` is false.

add-on/panels/test/panel_test.js
```diff
@@ -1475,4 +1478,129 @@ describe("loop.panel", function() {
...
+        }
+      });
+
+      TestUtils.Simulate.click(overlay);
+
+      expect(view.state.shouldOpenSharePanel).eql(false);
+      expect(panel.classList.contains("share-panel-open")).eql(false);
+      sinon.assert.calledOnce(dispatcher.dispatch);
+      sinon.assert.calledWithExactly(dispatcher.dispatch,
+        new sharedActions.OpenRoom({ roomToken: "fakeToken" }));
```
Standard8: The `sinon.assert`s here are duplicating the `OpenRoom` test. We should only need to test that there.

add-on/panels/test/panel_test.js
```diff
@@ -1475,4 +1478,129 @@ describe("loop.panel", function() {
...
+    });
+
+    it("should close the share panel when clicking outside the panel", function() {
```
Standard8, line 1651: We also need to test for closing the Hello panel.

Standard8:
For some reason, testing locally, I've not been able to get the share panel to disappear once the focus is lost/the room has been opened. Can you retest that please and see if you can reproduce it?

Also, please can you rebase on the latest master and check the functional tests - I believe the change in flow here is going to break the test.

add-on/panels/css/desktop.css
```diff
@@ -0,0 +1,215 @@
+/* Room share panel */
```
Standard8, line 10: Please add a license version block to the top of this file - you can use the same ones as for desktopViews.jsx

add-on/panels/js/panel.jsx
```diff
@@ -984,6 +976,119 @@ loop.panel = (function(_, mozL10n) {
...
+      this.setState(this.props.store.getStoreState());
+    },
+
+    onDocumentHidden: function() {
+      this.state.showPanel && this.handleClosePanel();
+    },
+
+    componentWillUpdate: function(nextProps, nextState) {
+      // If we've just created a room, close the panel - the store will open
+      // the room.
```
Standard8: The comment here doesn't seem to make sense wrt the code - the code is getting the panel to open, whereas this is talking about closing the panel.

ui/ui-showcase.jsx
```diff
@@ -760,6 +768,19 @@
 
             <FramedExample cssClass="fx-embedded-panel"
                            dashed={true}
+                           height={250}
+                           summary="Share panel"
```
Standard8, line 772: If in fake-mozLoop.js you add this in the GetLoopPref function, the facebook button should display as well:

``` js
        case "facebook.enabled":
          return true;
```

add-on/panels/css/desktop.css
```diff
@@ -0,0 +1,215 @@
...
+.share-panel-open > .room-invitation-overlay {
+  transform: translateX(0);
+}
+
+.share-panel-open > .share-panel-overlay {
+  display: block;
+}
+
+.share-panel-container > .room-invitation-overlay > .room-invitation-content > .room-context-header {
+  text-align: left;
```
Standard8, line 247: This needs an rtl version.

add-on/panels/css/desktop.css
```diff
@@ -0,0 +1,215 @@
...
+  cursor: pointer;
+  height: 34px;
+  border-radius: 4px;
+  background-color: #ebebeb;
+  display: flex;
+  flex-wrap: nowrap;
+  justify-content: center;
+  align-items: center;
+  flex-grow: 1;
+  margin-right: 20px;
```
Standard8, line 121: The margin-right needs to be reset for rtl mode with a margin-left added.

add-on/panels/test/desktopViews_test.js
```diff
@@ -0,0 +1,368 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
```
Standard8: nit: please use the public domain version for test files.

add-on/panels/test/panel_test.js
```diff
@@ -1561,4 +1565,124 @@ describe("loop.panel", function() {
...
+          hidden: true
+        }
+      });
+
+      expect(view.state.showPanel).eql(false);
+      sinon.assert.calledOnce(view.closeWindow);
+    });
+
+    describe("#openRoom", function() {
+      it("should open the room and close the panel", function() {
```
Standard8: This should be part of the main section where we're testing closing the panel. Something like change the visibility & check that `openRoom` is called - note that this doesn't (and shouldn't) test the close panel any more.

PR #168, opened 2016-02-10
Bug 1246934 - Report the add-on version to the server.

Standard8:


add-on/chrome/bootstrap.js
```diff
@@ -29,6 +29,9 @@ XPCOMUtils.defineLazyModuleGetter(this, "Task",
  * to provide the required loop functions for the window.
  */
 var WindowListener = {
+  // Records the add-on version once we know it.
+  addonVersion: "unknown",
```
mikedeboer, line 33: Why tack it to `WindowListener` and not as a global like you did in MozLoopService.jsm?

Standard8, line 33: I generally prefer to avoid global variables, and tack them to where they are used. MozLoopService is different due to the frozen interfaces.

PR #171, opened 2016-02-11
Bug 1247424 - Import l10n changes from m-c and get git-export working again

Standard8:


Makefile
```diff
@@ -349,8 +356,10 @@ dist_export:
 	$(RSYNC) $(ADD-ON)/install.rdf.in $(DIST_EXPORT_DIR)
 	# jar.mn is used for Firefox build
 	rm -f $(DIST_EXPORT_DIR)/chrome.manifest
-	# Add the jar file.
+	rm -f $(DIST_EXPORT_DIR)/chrome/locale/chrome.manifest
+	# Add the jar files.
 	$(RSYNC) $(ADD-ON)/jar.mn $(DIST_EXPORT_DIR)
+	$(RSYNC) locale/jar.mn $(DIST_EXPORT_DIR)/chrome/locale
```
Mardak, line 366: Do we need to generate what would be at `browser/extensions/loop/chrome/locale/moz.build` equivalent? Or that'll just be checked in to m-c? Although I suppose we don't generate `browser/extensions/loop/moz.build` either.

Standard8, line 366: The moz.build files are excluded from the export process. Generally they don't contain specific info about our files, so I didn't really want to include them in our repo.

bin/localeUtils.py
```diff
@@ -0,0 +1,29 @@
...
+# Utility module to provide utilities for dealing with locales.
+##
+
+import os
+
+
+def getLocalesList(src_dir):
+    def convertLocale(locale):
+        # Convert loop-client-l10n repo names to loop repo names.
+        return locale.replace('_', '-').replace('templates', 'en-US')
```
Mardak: We can move out `convertLocale` which is also used in `bin/locale_update.py`

bin/generateChromeManifest.py
```diff
@@ -65,11 +61,7 @@ def main(l10n_src, input_file_name, output_file_name):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
-        description="Loop Stand-alone Client localization update script")
-    parser.add_argument("--src",
-                        default=DEF_L10N_SRC,
-                        metavar="path",
-                        help="Source path for l10n resources. Default = " + DEF_L10N_SRC)
+        description="Loop chrome manifest generation script")
```
Mardak, line 62: Oh, we should probably update the description for `locale_update.py` too.

bin/generateLocaleManifest.py
```diff
@@ -0,0 +1,65 @@
...
+
+# defaults
+DEF_L10N_SRC = os.path.join("locale")
+DEF_INPUT_FILE_NAME = os.path.join("locale", os.extsep.join(["jar", "mn"]))
+DEF_OUTPUT_FILE_NAME = os.path.join("built", "add-on", "chrome", "locale",
+                                    os.extsep.join(["chrome", "manifest"]))
+
+
+def main(l10n_src, input_file_name, output_file_name):
+    print("Using l10n tree from", l10n_src)
```
Mardak: We can drop the `print` and `import` (also in `generateChromeManifest`) as these aren't that useful given how these are run from `Makefile`.

Makefile
```diff
@@ -270,7 +270,10 @@ standalone: node_modules $(built_standalone_js_files) $(built_standalone_shared_
 	$(RSYNC) shared $(BUILT)/$@/content
 
 .PHONY: add-on
-add-on: node_modules $(built_add_on_js_files) $(built_add_on_shared_js_files) $(built_add_on_l10n_files) $(BUILT)/$(ADD-ON)/chrome.manifest $(BUILT)/add-on/chrome/content/preferences/prefs.js
+add-on: node_modules $(built_add_on_js_files) $(built_add_on_shared_js_files) \
+	      $(built_add_on_l10n_files) $(BUILT)/$(ADD-ON)/chrome.manifest \
+	      $(BUILT)/$(ADD-ON)/chrome/locale/chrome.manifest \
+	      $(BUILT)/add-on/chrome/content/preferences/prefs.js
```
Mardak, line 279: Let's make these dependencies all be on separate lines as we can't get it to fit on a single line.

bin/generateLocaleManifest.py
```diff
@@ -0,0 +1,65 @@
...
+
+    input_file = io.open(input_file_name, "r")
+
+    jar_mn = input_file.read()
+
+    # Drop any line that doesn't begin with %.
+    manifest_output = "\n".join(line for line in jar_mn.split("\n") if (re.match("^%", line)))
```
Mardak: Do we expect anything other than the one line of `% locale loop @AB_CD@ %locale/@AB_CD@/` ? If not, we could skip all the reading/replacing and just generate the manifest file with desired `locale` registrations.

bin/generateChromeManifest.py
Standard8, line 62: That's `Loop Stand-alone Client localization update script` which I think is fine already.

Mardak, line 62: Well, it updates both the add-on and standalone localization files. ;)

bin/generateLocaleManifest.py
Standard8: Yeah, you're right, I think that would simplify it a bit.

bin/generateChromeManifest.py
Standard8, line 62: Ah, good catch!

PR #183, opened 2016-02-15
Bug 1245147 - Inform the link clicker when tab sharing is paused

mancas:


shared/css/conversation.css
```diff
@@ -376,6 +376,34 @@ html[dir="rtl"] .settings-menu.dropdown-menu {
...
+  box-shadow: 0 2px 4px rgba(0,0,0,.25);
+  display: inline-block;
+  margin: auto;
+  padding: 10px 15px;
+}
+
+.focus-stream > .room-inner-info-area > .remote-stream-paused > h1 {
+  font-size: 16px;
+  line-height: 30px;
+  padding-left: 42px;
```
mikedeboer, line 392: When you're using pixel values for font-related properties, instead of em/ rem, can you add a comment above them to explain why? My guess is that we can use em/ rem values here as well...

shared/js/views.jsx
```diff
@@ -796,6 +797,18 @@ loop.shared.views = (function(_, mozL10n) {
         return <div className="no-video"/>;
       }
 
+      if (this.props.shareCursor && this.props.screenSharingPaused) {
+        return (
+          <div className="room-inner-info-area">
```
Standard8: I'm not really happy with this here. Although we're not displaying the stream, I don't think we should be randomly tearing down the video element just as sharing is paused. Additionally, it feels more standalone specific at the moment, as it'll only be displayed on the standalone?

Could we instead move this to be handled in standalone in the StandaloneRoomInfoArea ? It might still need a display:none on the video element.

If its very difficult or doesn't make sense, I'll be happy to reconsider, but I'd like to see an attempted prototype or explanations first.

fcampo:
Could you add a check on [line 757](https://github.com/mozilla/loop/pull/183/files#diff-b96108efaa2799deaa24c8391d34a510R757)??

```
if (this.props.shareCursor && !this.props.screenSharingPaused) {
  // add listeners
}

```

it would help to stop sending the events to link generator when it's paused, as per [bug 1250166](https://bugzilla.mozilla.org/show_bug.cgi?id=1250166)

shared/js/activeRoomStore.js
```diff
@@ -151,6 +151,8 @@ loop.store.ActiveRoomStore = (function(mozL10n) {
         roomInfoFailure: null,
         // The name of the room.
         roomName: null,
+        // True when sharing screen has been paused.
+        streamPaused: false,
```
Standard8, line 156: This needs adding to the list of `_statesToResetOnLeave` - otherwise leaving a room whilst in the pause state, and then re-joining it, does the wrong thing.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -783,6 +802,7 @@ loop.standaloneRoomViews = (function(mozL10n) {
     },
 
     render: function() {
+      var screenSharingPaused = this.props.activeRoomStore.getStoreState("streamPaused");
```
Standard8: The activeRoomStore state is already included in the view - see `getInitialState`/`componentWillMount` - all you need to do here is `this.state.streamPaused` (and you can drop the intermediate variable.

shared/css/conversation.css
```diff
@@ -376,6 +376,34 @@ html[dir="rtl"] .settings-menu.dropdown-menu {
...
+  margin: auto;
+  padding: 10px 15px;
+}
+
+.focus-stream > .room-inner-info-area > .remote-stream-paused > h1 {
+  font-size: 1.6rem;
+  line-height: 3rem;
+  padding-left: 42px;
+  position: relative;
+  text-align: left;
```
Standard8, line 394: These classes will need rtl versions of them specifying.

Please also add a paused view to the ui-showcase so that we can easily view it and inspect for differences.

shared/js/views.jsx
```diff
@@ -920,7 +922,8 @@ loop.shared.views = (function(_, mozL10n) {
 
       var screenShareStreamClasses = classNames({
         "screen": true,
-        "focus-stream": this.props.displayScreenShare
+        "focus-stream": this.props.displayScreenShare,
+        "screen-sharing-paused": this.props.screenSharingPaused
```
Standard8, line 926: Please add a unit test to check the new class is added correctly.

standalone/test/standaloneRoomViews_test.js
```diff
@@ -890,6 +890,20 @@ describe("loop.standaloneRoomViews", function() {
         });
       });
 
+      describe("Screen sharing paused", function() {
+        it("should display paused view if the screen share has been stopped", function() {
+          activeRoomStore.setStoreState({
+            roomState: ROOM_STATES.HAS_PARTICIPANTS,
+            streamPaused: true
+          });
+          clock.tick(loop.standaloneRoomViews.StandaloneRoomInfoArea.RENDER_WAITING_DELAY);
```
Standard8: You don't need the clock.tick here - the message is displayed straight away (its only certain cases we have a delay).

PR #185, opened 2016-02-16
Bug 1244721 - Make panel gear icon bigger

mancas:


Mardak:
There's some style rules that seem obsolete now, e.g., `-moz-margin-start: .5em;` used to give some space before the button because "sign in/up" link was directly adjacent.

PR #186, opened 2016-02-16
Bug 1247785 - Remove unused strings for Invite Text header

mancas:


mancas:
@chrafuse I've asked @Standard8 on irc and we just need to remove the strings from en-US.

chrafuse:
Looks good. r=chrafuse

PR #187, opened 2016-02-16
Bug 1245150 - Change state at which to trigger StartBrowserShare action

piatra:
Previously StartBrowserShare was dispatched when state changed to room
joined (this is seconds later after the conversation window had already opened).
The issue was that the conversation window and the share tab infobar
appeared at different moments of time.

---

Here is a gif of the infobar action in progress

[gif of solution](http://i.imgur.com/VoieEox.gifv)

I've added the event dispatch at the soonest moment it made sense, after media access has been granted just before joining the room. It is better synchronized with the conversation window that popups up. Are there any concerns about those events? Is there a scenario when the application might not go through those 2 states?

### DIFF
--- a/add-on/chrome/bootstrap.js
+++ b/add-on/chrome/bootstrap.js
@@ -543,7 +543,11 @@ var WindowListener = {
           gBrowser.addEventListener("click", this);
         }
 
-        this._maybeShowBrowserSharingInfoBar();
+        if (this._roomSessionConnected) {
+          this._maybeShowBrowserSharingInfoBar();
+        } else {
+          this._showBrowserLoadingInfoBar();
+        }
 
         // Get the first window Id for the listener.
         let browser = gBrowser.selectedBrowser;
@@ -567,6 +571,7 @@ var WindowListener = {
        */
       roomSessionConnected: function() {
         this._roomSessionConnected = true;
+        this._maybeShowBrowserSharingInfoBar();
       },
 
       /**
@@ -717,12 +722,25 @@ var WindowListener = {
         return str;
       },
 
+      _showBrowserLoadingInfoBar: function() {
+        let box = gBrowser.getNotificationBox();
+        let pendingBtnString = "Connecting...";
+        let bar = box.appendNotification(
+          pendingBtnString,
+          kBrowserSharingNotificationId,
+          // Icon is defined in browser theme CSS.
+          null,
+          box.PRIORITY_WARNING_LOW);
+      },
+
       /**
        * Shows an infobar notification at the top of the browser window that warns
        * the user that their browser tabs are being broadcasted through the current
        * conversation.
        */
       _maybeShowBrowserSharingInfoBar: function() {
+        // Hide connecting info bar
+        this._hideBrowserSharingInfoBar();
         // Pre-load strings
         let pausedStrings = {
           label: this._getString("infobar_button_restart_label2"),
@@ -736,7 +754,7 @@ var WindowListener = {
         };
         let initStrings = this._browserSharePaused ? pausedStrings : unpausedStrings;
         // XXX This is work in progress.
-        let pendingBtnString = "Connecting...";
+        let box = gBrowser.getNotificationBox();
         let bar = box.appendNotification(
           initStrings.message,
           kBrowserSharingNotificationId,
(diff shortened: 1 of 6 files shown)


mikedeboer:
Well, yes, this looks reasonable and much more like the experience we'd like to have, but I'm curious: what happens when you pause/ stop sharing _before_ the stream is actually shared? My guess: the state gets whacked.

I'd be more comfortable with this change if you found a way to disable the buttons inside the infobar whilst the stream is connecting.

piatra:
That makes sense. I'll look into it, thanks.

piatra:
I've added a new action `RoomSessionConnected` in order to signal when connection has been achieved.

Between the moment the notification bar appears and session is established the pause/resume button will not do anything (disconnect button still works as before).

The thing left to do (as far as I can see) is have some way of signaling to the user that the pause/resume button is inactive until that status is reached. I could add `disabled=true` to the button but the difference is hardly noticeable, another way would be to change button strings to `Connecting` and then to `Pause or Resume` more obvious imo. 

I would like some feedback on this solution before going forward with it + writing tests. @mikedeboer

add-on/panels/js/roomViews.jsx
```diff
@@ -401,9 +401,15 @@ loop.roomViews = (function(mozL10n) {
       }
 
       // Automatically start sharing a tab now we're ready to share.
+      if (this.state.roomState === ROOM_STATES.MEDIA_WAIT &&
+          nextState.roomState === ROOM_STATES.JOINING) {
+        this.props.dispatcher.dispatch(new sharedActions.StartBrowserShare());
```
Mardak, line 196: Does this successfully share the tab? I believe dispatching this action eventually passes the window to the sdk, but if the sdk isn't set up per `SESSION_CONNECTED` state, I would expect tab sharing to fail.

add-on/chrome/bootstrap.js
```diff
@@ -671,10 +683,17 @@ var WindowListener = {
...
+            label: this._roomSessionConnected ? initStrings.label : pendingBtnString,
             accessKey: initStrings.accessKey,
             isDefault: false,
             callback: (event, buttonInfo, buttonNode) => {
+              // Prevent any action before connecting to a room.
+              if (!this._roomSessionConnected) {
+                // You must throw an error to prevent the notification bar
+                // from closing when the button is clicked.
+                throw new Error('Prevent notification bar close');
+              }
```
Mardak: We should get ui feedback before going too far here. I would think we would want to change the message (not button) text and perhaps not bother showing the "pause" button.

add-on/chrome/modules/MozLoopAPI.jsm
```diff
@@ -192,6 +192,16 @@ const kMessageHandlers = {
   },
 
   /**
+   * Triggered at ROOM_STATES.SESSION_CONNECTED meant to notify the connection
+   * status.
+   */
+  RoomSessionConnected: function(message, reply) {
+    let win = Services.wm.getMostRecentWindow("navigator:browser");
+    win.LoopUI.roomSessionConnected();
```
Mardak, line 208: We should probably have a better way to make sure we call `roomSessionConnected` to the correct window, but I see this pattern elsewhere in this file, so this is fine for now.

add-on/chrome/bootstrap.js
piatra: The error is thrown so that the button will not close the window 
https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Method/appendNotification
The API specifies that.
Changing the UI (from a disabled button to an active one) doesn't help much (difference is very small) it is barely noticeable. Showing the button after the connection is established might look odd and editing the buttons in the notification bar is potentially difficult. You would need a reference to the notification bar created (please correct me if I'm wrong).

add-on/panels/js/roomViews.jsx
piatra, line 196: I wasn't aware of that, it just worked for me when I tested. I will look into it more.

Standard8, line 196: Can you also make sure this works in the rejoin case? I'm thinking when the rejoin button is hit.

piatra:
@Mardak I updated the branch and you should be able to test it now w/ FF Beta.

I updated the notifications such that now the bar immediately shows up but only displaying a "Connecting..." message. 
Once `this._roomSessionConnected = true;` the notification is removed and the usual notification bar is added.

Tested this with reconnecting and works as usual. What do you think about this solution?

Mardak:
This bug was wontfixed now that https://github.com/mozilla/loop/pull/463 merged.

PR #191, opened 2016-02-17
Bug 1248960 - Add an option to run the dev server without watching files, and switch functional tests to use it, and update the url for the tests.

Standard8:


Makefile
```diff
@@ -435,7 +435,11 @@ test: lint karma
 
 .PHONY: runserver
 runserver: remove_old_config
-	$(REPO_BIN_DIR)/run-server.sh
+	@if [[ -z "${NO_WATCH}" ]]; then \
+	  $(REPO_BIN_DIR)/run-server.sh; \
+	else \
+	  node bin/server.js; \
+	fi
```
Mardak: Might be cleaner to just have a separate make target, e.g., `runserver-only` or `runserver-nowatch` instead of environment variables. Could be useful and easier to run separately.

Standard8: yeah, lets go with that - I was originally thinking of keeping `make runserver` so that is was compatible with loop-client, but I can just add another target there whilst we transition.

Standard8:
@Mardak updated PR.

PR #196, opened 2016-02-18
Bug 1229195 - prevent a JS error related to the UITour on beta and all…

mikedeboer:
…ow for a longer period of time before timeout in mochitests. r=Standard8


add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -981,7 +981,11 @@ var MozLoopServiceInternal = {
             if (kEventNamesMap[eventName]) {
               eventName = kEventNamesMap[eventName];
 
-              UITour.clearAvailableTargetsCache();
+              if (UITour.clearAvailableTargetsCache) {
+                UITour.clearAvailableTargetsCache();
```
Standard8, line 987: It looks like this if statement got lost in https://github.com/mozilla/loop/commit/36af5ef9c60669695631127864cc0f40dd57174a#diff-86d410f7f808338d1201d6c62815c743L933
Please can you add back the comment about which versions it applies to so that we'll be able to clean it up later more easily.

mikedeboer:
Merged by 67a989ff8d86c0b06f5c6b6a8663072daa4d7b8f

PR #205, opened 2016-02-23
Bug 1248895 - Open Hello panel after FTU is closed

daveccrit:
r= @Standard8 

Should we setup MochiTests for this as well? The slideshow doesn't have tests yet, but I feel this should have a mochitest added for it at some point. Maybe we can pair up to write a mochitest for this tomorrow PST morning sometime


Standard8:
@daveccrit fyi looks like you forgot to update the PR branch before pushing to master, so I've closed this.

daveccrit:
yep, this was the one I mentioned yesterday about forgetting to add the
r=standard8 to the commit message

PR #206, opened 2016-02-23
Bug 1245710 - link clicker intro screen

chrafuse:
@dmose First pass of working link clicker introduction message.


chrafuse:
@dmose Added join room to close button (was that decided with Holly?) and Ok, Got It! button.
Fixed Travis CI issues.
Changed how you set introSeen from the introSeen property and overrides the localStorage setting for testing purposes and UI showcase.
http://localhost:3000/ui/#standalone-room-introduction-message

dmose:
One issue I see is that the text seems to be missing a word, to the point where sentence doesn't quite make sense.  "just don't forget to let use your camera and mic"

dmose:
I also saw the slideshow on both the first and second times I made a Hello call, for some reason.

dmose:
This wants a high-priority spin-off bug for unit tests.

standalone/content/css/webapp.css
```diff
@@ -8,6 +8,7 @@ body,
   /* Required for `justify-content: space-between` to divide space equally
    * based on the full height of the page */
   height: 100%;
+  min-height: 100%;
```
dmose: Try to find out why this is necessary and a comment with it.

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
   background: #000;
 }
 
+#main {
+  position: relative;
+}
```
dmose: This doesn't seem to be necessary; let's get rid of it.  Even if something goes wrong, I don't think this is where we'd want to put the relative.

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
...
+
+.intro-overlay {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 1999;
+  background-color: rgba(0,0,0,0.5); /*dim the background*/
```
dmose: Spaces on between comment delimiters and text, please.

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
...
+}
+
+/* Intro Overlay Content */
+.slideshow {
+  position: relative;
+  -moz-user-select: none;
+  cursor: default;
+  height: 450px;
+  width:  620px;
+  text-align: left;
```
dmose, line 45: Unnecessary whitespace in the last two lines.

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
...
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
dmose, line 49: Add a comment that the above three lines are for centering.

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
...
+  padding-left: 50px;
+  padding-bottom: 20px;
+  width: 600px;
+}
+
+.slide-layout > h2 {
+  font-family: sans-serif;
+  font-size: 2.8rem;
+  font-weight: 600;
+  white-space: normal;
```
dmose, line 102: Unless we're overriding something, normal should be the default value here.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -543,12 +544,24 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
     },
-
     getInitialState: function() {
+      // test that it gets set -  remove after testing - uncomment next line to reset
+      // localStorage.removeItem("introSeen");
+      var introSeen = false;
+      if (this.props.introSeen !== undefined) {
+        if (this.props.introSeen) {
+          introSeen = true;
+        }
```
dmose, line 619: Add a comment about why the above is necessary.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -788,6 +802,84 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
+      text: React.PropTypes.string.isRequired,
+      title: React.PropTypes.string.isRequired
+    },
+
+    handleGotItClick: function(event) {
+      event.stopPropagation();
+      event.preventDefault();
+      var overlayElem = document.getElementsByClassName("intro-overlay")[0];
+      overlayElem.classList.add("hide");
+      localStorage.setItem("introSeen", true);
```
dmose: This should be a string.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -722,7 +735,7 @@ loop.standaloneRoomViews = (function(mozL10n) {
     render: function() {
       var displayScreenShare = !!(this.state.receivingScreenShare ||
         this.props.screenSharePosterUrl);
-
+      var overlay = (!this.state.introSeen) ? <IntroOverlayView joinRoom={this.joinRoom} /> : false;
```
dmose: s/false/null to match prevailing style.  Also, please invert the test for readability.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -788,6 +802,84 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
+      imageClass: React.PropTypes.string.isRequired,
+      indexClass: React.PropTypes.string.isRequired,
+      joinRoom: React.PropTypes.func.isRequired,
+      text: React.PropTypes.string.isRequired,
+      title: React.PropTypes.string.isRequired
+    },
+
+    handleGotItClick: function(event) {
+      event.stopPropagation();
+      event.preventDefault();
```
dmose, line 881: Blank line here for readability, plz.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -788,6 +802,84 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
+  var Slide = React.createClass({
+    propTypes: {
+      imageClass: React.PropTypes.string.isRequired,
+      indexClass: React.PropTypes.string.isRequired,
+      joinRoom: React.PropTypes.func.isRequired,
+      text: React.PropTypes.string.isRequired,
+      title: React.PropTypes.string.isRequired
+    },
+
+    handleGotItClick: function(event) {
```
dmose, line 879: Please move this function to the component containing the overlay element and pass it in as a callback.  The callback (in addition to the stuff in this function) simply should change this.state (or dispatch a flux action if that's appropriate) so that it re-renders without the overlay element.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -788,6 +802,84 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
+      );
+    }
+  });
+
+  var IntroOverlayView = React.createClass({
+    propTypes: {
+      joinRoom: React.PropTypes.func.isRequired
+    },
+
+
```
dmose: Extra blank line here.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -788,6 +802,84 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
+    handleCloseClick: function(event) {
+      event.stopPropagation();
+      event.preventDefault();
+      // hide or remove the
+      var overlayElem = document.getElementsByClassName("intro-overlay")[0];
+      overlayElem.classList.add("hide");
+      this.props.joinRoom();
+    },
```
dmose, line 928: I believe the semantics here should be identical to the "Got it" button (ie mark the slideshow as seen).  Feel free to check with Holly to be sure.  So this can probably share the handler with the Got It button.

standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -788,6 +802,84 @@ loop.standaloneRoomViews = (function(mozL10n) {
...
+
+    render: function() {
+      var slideNode = {
+        id: "slide1",
+        imageClass: "slide1-image",
+        title: mozL10n.get("intro_slide_title"),
+        text: mozL10n.get("intro_slide_copy", {
+          clientSuperShortname: mozL10n.get("clientSuperShortname")
```
dmose, line 936: Excitingly, clientSuperShortname is currently only defined in add-on.properties for the browser add-on, rather than in shared.properties for everyone.  So we technically _have_ the translated string we need, but it's not in the right places... Hmmm.  One thing that comes to mind is something like a sed script, assuming all the translations are already done (are they?).  We could also choose to package up add-on.properties with the standalone temporarily.  @Standard8: thoughts?

dmose:
This is almost there with a few key tweaks and some more minor ones.  feedback+  :-)

standalone/content/css/webapp.css
```diff
@@ -8,6 +8,7 @@ body,
   /* Required for `justify-content: space-between` to divide space equally
    * based on the full height of the page */
   height: 100%;
+  /*min-height: 100%;*/
```
dmose: Just kill this rather than commenting it out.

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
...
+}
+
+.intro-overlay {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 1999;
+  background-color: rgba(0,0,0,0.5); /* dim the background */
```
dmose: Still need spaces here.

dmose: (in the rgba args too)

standalone/content/js/standaloneRoomViews.jsx
dmose, line 879: In addition to spinning off a new bug for this, please add XXX comments around the relevant pieces that will be called back so that future code readers will get relevant ideas about what to do with the code.

standalone/content/js/standaloneRoomViews.jsx
dmose, line 936: OK, we've dealt with this.

standalone/content/js/standaloneRoomViews.jsx
dmose, line 881: This still needs to happen.

locale/en-US/standalone.properties
```diff
@@ -61,3 +61,5 @@ rooms_read_while_wait_offer=Want something to read while you wait?
 ## replaced by the brand name.
 standalone_title_with_room_name={{roomName}} — {{clientShortname}}
 status_error=Something went wrong
+
+clientSuperShortname=Hello
```
dmose, line 67: Please add it like this:

```
## LOCALIZATION NOTE(clientSuperShortname): This should not be localized and
## should remain "Hello" for all locales.
clientSuperShortname=Hello
```

standalone/content/css/webapp.css
```diff
@@ -25,6 +26,130 @@ body,
   background: #000;
 }
 
+#main {
+}
```
dmose: This can go, I think, since it has no rules.  :-)

chrafuse:
@dmose Created code improvement bug and is dependent on this bug.  Addressed comments.

dmose:
Superceded by PR #222.

PR #213, opened 2016-02-25
No Bug - Timing change within the check video and switch to chatbox to allow DOM catch up. r=standard8

daveccrit:
r? @Standard8


mozilla-autolander-deprecated:

Standard8:
r=Standard8

PR #218, opened 2016-02-29
Bug 1245808 - Re-enable browser_mozLoop_chat.js for e10s mode now that the issues with it seem to have gone away.

Standard8:


mikedeboer:
r=me

PR #225, opened 2016-03-01
Bug 1252450 - Enable the facebook button for shipped builds.

Standard8:


add-on/preferences/prefs.js
```diff
@@ -33,11 +33,7 @@ pref("loop.CSP", "default-src 'self' about: file: chrome:; img-src * data:; font
 pref("loop.fxa_oauth.tokendata", "");
 pref("loop.fxa_oauth.profile", "");
 pref("loop.support_url", "https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/cobrowsing");
-#ifdef LOOP_BETA
 pref("loop.facebook.enabled", true);
```
fcampo, line 36: is there any reason to keep the pref if we're not changing the value (or checking it) at any point in the code?

Standard8, line 36: It is checked in conversation.jsx (look for `facebook.enabled`), and then passed through to `DesktopRoomInvitationView`.

Its still a little new and hasn't been fully released yet, so I was going on the conservative side to keep it in for a bit longer yet, in case it needed backing out for any reason.

fcampo, line 36: oh, I see it now on the invitation view. 
Still, we're checking a pref that is always `true`, which is weird, plus we'll need another bug eventually to get rid of the rests of it. We could remove it all, and back it up the whole thing if needed, instead of splitting it.

Anyway, if we want to go conservative, I'm happy with it as long as we don't forget about the remaining code (followup?)

Standard8, line 36: I'd rather keep it simple to disable if we need to. It isn't that much extra code compared to the rest and it shouldn't hurt performance.

I've filed [bug 1252830](https://bugzilla.mozilla.org/show_bug.cgi?id=1252830) on removing it later on.

PR #228, opened 2016-03-02
Bug 1252061 - The clicker UI tile needs to be moved to the bottom rig…

mancas:
…ht of the page


shared/css/conversation.css
```diff
@@ -908,8 +908,8 @@ body[platform="win"] .share-service-dropdown.overflow > .dropdown-menu-item {
 }
 
 .text-chat-disabled > .text-chat-entries {
-  /* When text chat is disabled, the entries box should be 100% height. */
-  height: 100%;
+  /* 220px is the heifht of the ads tile. */
```
fcampo: nit: height

fcampo: Maybe more details on the comment, like 
`/* When text chat is disabled, we want to show a tile ad with 220px height */`

standalone/content/css/webapp.css
```diff
@@ -586,32 +586,54 @@ html[dir="rtl"] .standalone-info-bar > .media-control-buttons > .btn-hangup:afte
...
+  margin: 8px 0;
+  text-transform: uppercase;
+}
+
+.ads-tile > .ads-wrapper > a > .room-waiting-help {
+  background: #ebebeb url("../shared/img/glyph-help-no-circle-16x16.svg") center center no-repeat;
+  background-size: 16px 16px;
+  border: solid 4px #fff;
+  border-radius: 50%;
+  height: 30px;
```
fcampo, line 617: I know that guidelines ask for alphabetical order, but it would be much easier to read if related constraints are together (height&width for size)

PR #229, opened 2016-03-02
No bug. Fix installation of the add-on in functional tests - install it permanently rather that temporary as we still need to restart.

Standard8:


mozilla-autolander-deprecated:

mancas:
r=me thanks!

PR #235, opened 2016-03-03
Bug 1252162 - Link clicker intro screen tests

chrafuse:
@dmose Created two tests that checks that introSeen can show and not show the intro message slide.

Could test if localStorage is set but had issue with simulate button click and overlay element becomes undefined.


daveccrit:
ship it

PR #236, opened 2016-03-03
Bug 1252088 - Ripple click effect needs more unit tests. r=Standard8

mancas:


mancas:
We already have tests for the _onRemoteCursorPositionChange in conversation_test.js

Standard8:
> We already have tests for the _onRemoteCursorPositionChange in conversation_test.js

Ah yes, I'd missed that.

Though I think we still need a test for `CursorClick` coming from the browser/LoopAPI similar to `#cursor position change`.

Can you also rebase on latest master so that travis passes please?

mancas:
I think you mean this test: https://github.com/mancas/loop/blob/bug1252088/shared/test/remoteCursorStore_test.js#L70 right?

Standard8:
> I think you mean this test: https://github.com/mancas/loop/blob/bug1252088/shared/test/remoteCursorStore_test.js#L70 right?

Yes, that's the one. Thanks for finding it!

PR #243, opened 2016-03-04
Bug 1252817 - Stop using the timeout for requests across the content/…chrome API as this interfers with network requests via the backend.

Standard8:


mikedeboer:
LGTM.

PR #244, opened 2016-03-04
Bug 1239972 - Adapt infobar message when link clicker in room or not

chrafuse:
@dmose first pass with test console.logs. Triggers on joining room and left room events.  Moved away from triggering from the startBrowserSharing in the bootstrap.js file.  UX needs to verify they want a different message when we are paused and unpaused in an empty room.  It is currently the same message but background colour is different.


chrafuse:
@dmose Addresses join and left room and switching tabs.

chrafuse:
@dmose Fixed accessKey for Stop sharing button.  2 commit, one removes the consoles.  We also have 2 strings that needs to be translated.

dmose:
Please fix the eslint error that caused travis to go red.

add-on/chrome/bootstrap.js
```diff
@@ -460,7 +460,6 @@ var WindowListener = {
           };
           this.playSound(options.sound);
         }
-
```
dmose, line 598: No need to get rid of this, I don't think.

add-on/chrome/bootstrap.js
```diff
@@ -681,28 +677,57 @@ var WindowListener = {
         return str;
       },
 
+      _browserSharePaused: false,
+      /**
+       * Set correct strings for infobar notification based on if paused or empty.
+       */
```
dmose, line 852: Please add jsdoc comments for the argument(s).

add-on/chrome/bootstrap.js
```diff
@@ -681,28 +677,57 @@ var WindowListener = {
         return str;
       },
 
+      _browserSharePaused: false,
```
dmose: Please add JSDoc for this too.

add-on/chrome/bootstrap.js
```diff
@@ -681,28 +677,57 @@ var WindowListener = {
...
+        let stopButtonAccessKey = this._getString("infobar_button_stop_accesskey");
+
+        let stopSharingMessage = this._getString("infobar_screenshare_stop_sharing_message2");
+        let stopSharingNoGuestMessage = this._getString("infobar_screenshare_stop_no_guest_message");
+
+        let sharingMessage = this._getString("infobar_screenshare_browser_message3");
+        let sharingNoGuestMessage = this._getString("infobar_screenshare_no_guest_message");
+
+        let infoStrings = {};
+        let sharePaused = this._browserSharePaused;
```
dmose: Let's pass in for more modularity.

add-on/chrome/bootstrap.js
dmose: Actually, rather than making this a member variable, make it locally scoped to _maybeShowBrowserSharingInfoBar using let.

add-on/chrome/bootstrap.js
dmose, line 852: As well as the semantic of how this._browserSharePaused is involved.

add-on/chrome/bootstrap.js
```diff
@@ -681,28 +677,57 @@ var WindowListener = {
...
+
+        let stopButtonLabel = this._getString("infobar_button_stop_label2");
+        let stopButtonAccessKey = this._getString("infobar_button_stop_accesskey");
+
+        let stopSharingMessage = this._getString("infobar_screenshare_stop_sharing_message2");
+        let stopSharingNoGuestMessage = this._getString("infobar_screenshare_stop_no_guest_message");
+
+        let sharingMessage = this._getString("infobar_screenshare_browser_message3");
+        let sharingNoGuestMessage = this._getString("infobar_screenshare_no_guest_message");
+
```
dmose: Can't most of the message strings be declared in the then/else clauses?

add-on/chrome/bootstrap.js
```diff
@@ -681,28 +677,57 @@ var WindowListener = {
...
+       * Set correct strings for infobar notification based on if paused or empty.
+       */
+
+      _setInfoBarStrings: function(nonOwnerParticipants) {
+
+        let restartButtonLabel = this._getString("infobar_button_restart_label2");
+        let restartButtonAccessKey = this._getString("infobar_button_restart_accesskey");
+
+        let stopButtonLabel = this._getString("infobar_button_stop_label2");
+        let stopButtonAccessKey = this._getString("infobar_button_stop_accesskey");
```
dmose: Please put these near the places they're used.

add-on/chrome/bootstrap.js
```diff
@@ -681,28 +677,57 @@ var WindowListener = {
...
-        };
-        let unpausedStrings = {
-          label: this._getString("infobar_button_stop_label2"),
-          accesskey: this._getString("infobar_button_stop_accesskey"),
-          message: this._getString("infobar_screenshare_browser_message2")
-        };
-        let initStrings =
-          this._browserSharePaused ? pausedStrings : unpausedStrings;
-
+      _maybeShowBrowserSharingInfoBar: function(nonOwnerParticipants) {
```
dmose: Please add JSDoc for the argument.

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -1348,17 +1348,26 @@ this.MozLoopService = {
...
+      // The participant that joined isn't necessarily included in room.participants (depending on
+      // when the broadcast happens) so concatenate.
+      let isOwnerInRoom = room.participants.concat(participant).some(p => p.owner);
+      let window = gWM.getMostRecentWindow("navigator:browser");
+      if (participant.owner) {
+        // is joining call and determine if other is here
+        let participantRoomCount = room.participants.concat(participant).length;
+        window.LoopUI._maybeShowBrowserSharingInfoBar(participantRoomCount > 1);
+      } else {
+        if (isOwnerInRoom) {
```
dmose: Please do the simplification if just computing participant room count at the top, and don't check the owner.

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -1348,17 +1348,26 @@ this.MozLoopService = {
     LoopRooms.on("update", onRoomsChange);
     LoopRooms.on("delete", onRoomsChange);
     LoopRooms.on("joined", (e, room, participant) => {
+      // The participant that joined isn't necessarily included in room.participants (depending on
+      // when the broadcast happens) so concatenate.
+      let isOwnerInRoom = room.participants.concat(participant).some(p => p.owner);
```
dmose: I'm concerned about possible concatenation side-effects without cloning if room is a reference.  

Please separate this out into its own getter, with a big XXX describing why we're not using the existing roomsParticipantsCount.

Mardak:
There's an existing observer in bootstrap.js that listens for `loop-status-changed` that calls         `this.updateToolbarState(data);` that eventually checks:

https://github.com/mozilla/loop/blob/master/add-on/chrome/bootstrap.js#L390

```
       } else if (this.MozLoopService.roomsParticipantsCount > 0) {
          state = "active";
          this.roomsWithNonOwners().then(roomsWithNonOwners => {
            if (roomsWithNonOwners.length > 0) {
              mozL10nId += "-participantswaiting";
```

Could we reuse that for this bug?

chrafuse:
@dmose Added Mochitests for pause button type and label message in paused and sharing states.

add-on/chrome/bootstrap.js
```diff
@@ -523,13 +523,27 @@ var WindowListener = {
       },
 
       /**
+       * Set the current room token
+       *
+       * Push message parameters:
+       * - {String} roomToken  The current room that the link generator is connected to.
```
dmose: Since we're using null as a parameter as well, please document the semantics of what it means to have a roomToken set to null that here.  http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters.

add-on/chrome/bootstrap.js
```diff
@@ -523,13 +523,27 @@ var WindowListener = {
...
+       */
+      setCurrentRoomToken: function(roomToken) {
+        this._currentRoomToken = roomToken;
+      },
```
dmose: Please switch to the getter and setter syntax used elsewhere in this module.

dmose: Check out MozLoopService for some examples.

add-on/chrome/bootstrap.js
```diff
@@ -523,13 +523,27 @@ var WindowListener = {
...
+
+      /**
        * Start listening to selected tab changes and notify any content page that's
        * listening to 'BrowserSwitch' push messages.
        *
        * Push message parameters:
        * - {Integer} windowId  The new windowId for the browser.
        */
-      startBrowserSharing: function() {
+      startBrowserSharing: function(roomToken) {
```
dmose, line 650: Please document roomToken.

add-on/chrome/bootstrap.js
```diff
@@ -709,26 +725,55 @@ var WindowListener = {
...
+          infoStrings.accesskey = this._getString("infobar_button_stop_accesskey");
+        }
+        return infoStrings;
+      },
+
+      /**
+       * Indicates if tab sharing is paused.
+       * Set by tab pause button, startBrowserSharing and stopBrowserSharing.
+       * When link generator(owner) enters room we are sharing tabs by default.
+       * Set to default here as startBrowserSharing is called too late for
```
dmose: "Defaults to false"...

dmose: We need to check that this gets reset to false in all appropriate cases.

add-on/chrome/bootstrap.js
```diff
@@ -738,11 +783,11 @@ var WindowListener = {
           box.PRIORITY_WARNING_LOW,       // priority
           [{                              // buttons (Pause, Stop)
             label: initStrings.label,
-            accessKey: initStrings.accessKey,
+            accessKey: initStrings.accesskey,
             isDefault: false,
             callback: (event, buttonInfo, buttonNode) => {
               this._browserSharePaused = !this._browserSharePaused;
-              let stringObj = this._browserSharePaused ? pausedStrings : unpausedStrings;
+              let stringObj = this._setInfoBarStrings(this.LoopRooms.getNumParticipants(this.getCurrentRoomToken()) > 1, this._browserSharePaused);
```
dmose: Please use an intermediate variable to add to readability.

add-on/chrome/bootstrap.js
```diff
@@ -709,26 +725,55 @@ var WindowListener = {
...
+       * Set to default here as startBrowserSharing is called too late for
+       * infobar notification state.
+       */
+      _browserSharePaused: false,
+
+      /**
        * Shows an infobar notification at the top of the browser window that warns
        * the user that their browser tabs are being broadcasted through the current
        * conversation.
+       * @param  {Boolean} optional nonOwnerParticipants is true if a guest is present in the room
```
dmose: Please document what happens if this parameter is not passed.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -594,6 +595,16 @@ var LoopRoomsInternal = {
     return gGetAllPromise;
   },
 
+  getNumParticipants: function(roomToken) {
```
dmose, line 604: Please add jsdoc describing this function, and why it's necessary at all.

dmose, line 604: We need to explicitly define the error behavior here.

dmose, line 604: We also need xpcshell unit tests for the changes in this file; specifically one that fails without these changes and passes with it.

add-on/chrome/bootstrap.js
dmose: Please also document what happens if this._currentRoomToken is set to null.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -567,11 +568,11 @@ var LoopRoomsInternal = {
           eventEmitter.emit("delete", room);
           eventEmitter.emit("delete:" + room.roomToken, room);
         } else {
+          yield this.addOrUpdateRoom(room, !!orig);
+
```
dmose, line 571: We need to manually audit the (few, I hope) callers of this to make sure this change won't cause problems for them.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -556,6 +556,7 @@ var LoopRoomsInternal = {
       for (let room of roomsList) {
         // See if we already have this room in our cache.
         let orig = this.rooms.get(room.roomToken);
+        // let orig = extend({}, this.rooms.get(room.roomToken));
```
dmose: @Standard8, our recollection is that this was something you suggested relating to making sure we have the right version somehow.  Do recall what the point of using the commented-out form (which has some other issue) was?

add-on/chrome/modules/MozLoopAPI.jsm
```diff
@@ -188,8 +188,7 @@ const kMessageHandlers = {
     }
 
     let [windowId] = message.data;
-
-    win.LoopUI.startBrowserSharing();
+    win.LoopUI.startBrowserSharing(windowId);
```
dmose, line 194: I would suggesting changing windowId to roomToken, or at least documenting that we're depending on that being true.

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -1395,11 +1396,22 @@ this.MozLoopService = {
           message: localizedString,
           selectTab: "rooms"
         });
+
       }
```
dmose, line 1417: No need to add a blank line here, I don't think.

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -1374,13 +1373,15 @@ this.MozLoopService = {
...
 
       let window = gWM.getMostRecentWindow("navigator:browser");
+      if (window.LoopUI.getCurrentRoomToken() && window.LoopUI.getCurrentRoomToken() === room.roomToken) {
```
dmose: This might want a comment that this is to only show the infobar if this event is for the current room.

add-on/chrome/test/mochitest/browser_mozLoop_sharingListeners.js
```diff
@@ -157,13 +157,22 @@ add_task(function* test_infoBar() {
 
   let testBarProps = function() {
     let bar = getInfoBar();
-
```
dmose, line 166: No need for a change here.

add-on/chrome/test/mochitest/browser_mozLoop_sharingListeners.js
```diff
@@ -157,13 +157,22 @@ add_task(function* test_infoBar() {
...
+    // Pause button and type
     let button = bar.querySelector(".notification-button");
     Assert.ok(button, "There should be a button present");
+    Assert.equal(button.type, "pause", "The bar button should be type pause");
+
+    // Paused state and label message
+    button.click();
+    Assert.equal(bar.label, getLoopString("infobar_screenshare_stop_no_guest_message"), "The bar label should match when paused");
+    button.click();
   };
```
dmose: @Standard8, do you know how we could test connection and disconnection messages here too?

shared/js/activeRoomStore.js
```diff
@@ -832,7 +832,6 @@ loop.store.ActiveRoomStore = (function(mozL10n) {
         });
         return;
       }
-
```
dmose, line 839: No need for this change.

add-on/chrome/bootstrap.js
dmose: And please audit all callers that aren't passing in a parameter or executing with null set so be sure that something reasonable happens.  If it's possible to eliminate those callers, even better.

dmose:
A big step in the right direction; feedback+.

add-on/chrome/modules/LoopRooms.jsm
Standard8: We've just discussed this and determined that the copy of the room stored in `this.rooms` is not modified. So we don't need the `extend` function here.

add-on/chrome/test/mochitest/browser_mozLoop_sharingListeners.js
Standard8: I think the way to do this would be to use the `LoopRooms._setRoomsCache()` function (see `browser_LoopRooms_channel.js` for an example).

You could set a room list via that function. To actually trigger an infobar update, you could add an extra optional parameter which calls the checkParticipantsUpdate function or something like that.

chrafuse:
@Standard8, @dmose Fixed review issues started on test for guest present in room but was not taking.  any suggestions on the add-on/chrome/test/mochitest/browser_mozLoop_sharingListeners.js tests.  I may need to set up similar to add-on/chrome/test/mochitest/browser_LoopRooms_channel.js line 89.

dmose:
@chrafuse Note that some e10s changes to that test file that landed in the loop repo today, so it's conceivable that rebasing against the latest master will help, though I kinda suspect not.  If it doesn't, it'd be helpful to understand more about what kind of problems you're seeing...

dmose:
https://github.com/mozilla/loop/commit/62fe3f00fc69a28457408e9ce170897d7eca6921

dmose:
Is the commit with those e10s test changes.

dmose:
And code changes too, it seems....

chrafuse:
@Standard8, @dmose Latest code fixes review comments above and trying to fix mochitest or create test for connected guest state.  Not getting any mochi test output after Friday pull.  Any suggestion on that.  The infobar issues can be reviewed and is just the mochitest for connected quest not working.

chrafuse:
@Standard8 - Added infobar tests to browser_LoopRooms_channel.js, fails on no room found after setting up a room and using ROOM_TOKEN to match the roomToken generated but fails.  It seems it is required to now have a room in the rooms array exist to test the infobar at all.  Original tests fail now.  Any suggestions on how to connect or fake a rooms list with the containing room?

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -595,6 +596,24 @@ var LoopRoomsInternal = {
...
+   */
+  getNumParticipants: function(roomToken) {
+    MozLoopService.log.debug("bar roomToken", roomToken);
+    MozLoopService.log.debug("bar this.rooms", this.rooms);
+    MozLoopService.log.debug("bar this.rooms.has(roomToken)", this.rooms.has(roomToken));
+    if (this.rooms.has(roomToken)) {
+      return this.rooms.get(roomToken).participants.length;
+    }
+    // no room, log error and send back 0 to indicate none in room
+    throw new Error("No room found in current session.");
```
Standard8: I think you either want to change `_maybeShowBrowserSharingInfoBar` to check that the room token is valid before passing it to here and/or add a try catch, or you want to change this to return 0 and maybe log an error.

Standard8:
@chrafuse you'll also want to rebase against latest master - I've just landed a bustage fix for the browser_sharingTitleListeners.js test.

chrafuse:
@Standard8, @dmose
First run of browser_mozLoop_infobar.js

...
JavaScript error: file:///home/chrafuse/Documents/mozilla/gecko-dev/obj-x86_64-pc-linux-gnu/dist/bin/components/Weave.js, line 13: NS_ERROR_NOT_AVAILABLE: Component returned failure code: 0x80040111 (NS_ERROR_NOT_AVAILABLE) [nsIXPCComponents_Utils.import]
...
54 INFO TEST-UNEXPECTED-FAIL | browser/extensions/loop/chrome/test/mochitest/browser_mozLoop_infobar.js | Found an unexpected alert:alert at the end of test run - 
Completed ShutdownLeaks collections in process 27128
Completed ShutdownLeaks collections in process 27025
55 INFO TEST-START | Shutdown
56 INFO Browser Chrome Test Summary
57 INFO Passed:  32
58 INFO Failed:  1
59 INFO Todo:    0
60 INFO Mode:    e10s
61 INFO **\* End BrowserChrome Test Results ***

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -595,6 +595,27 @@ var LoopRoomsInternal = {
...
+      if (this.rooms && this.rooms.has(roomToken)) {
+        return this.rooms.get(roomToken).participants.length;
+      }
+      return 0;
+    }
+    catch (ex) {
+      // No room, log error and send back 0 to indicate none in room.
+      // throw new Error("No room found in current session: ", ex);
+      return 0;
+    }
```
dmose, line 615: XXX need to figure out the reasoning here and decide if this makes sense and also whether we should log something.

add-on/chrome/modules/MozLoopAPI.jsm
```diff
@@ -160,13 +160,15 @@ const kMessageHandlers = {
...
-   *                             {Number} windowId The window ID of the chat window
+   *                             {String} roomToken The room ID to start browser sharing and listeners.
    *                           ]
    * @param {Function} reply   Callback function, invoked with the result of this
    *                           message handler. The result will be sent back to
    *                           the senders' channel.
    */
   AddBrowserSharingListener: function(message, reply) {
+    // MozLoopService.log.debug("bar AddBrowserSharingListener message.data", message.data);
+
```
dmose: This wants to either be enabled and have a more understandable message, or removed.

chrafuse:
@dmose Added getNumParticipants tests to the test_looproom.js file.  Need to verify that the numbers are expected for before and after join.

add-on/chrome/modules/LoopRooms.jsm
dmose, line 615: This reasoning seems fine; I just uncommented the throw statement and turned it into an error.

dmose, line 615: By which I mean an error log message.

dmose:
The getNumParticipants tests look good to me.

add-on/chrome/bootstrap.js
```diff
@@ -635,16 +635,20 @@ var WindowListener = {
        * Start listening to selected tab changes and notify any content page that's
        * listening to 'BrowserSwitch' push messages.
```
dmose: I've updated this comment to describe the joined and left listener stuff too.

add-on/chrome/modules/MozLoopAPI.jsm
dmose: I've removed it.

locale/en-US/add-on.properties
```diff
@@ -203,7 +203,8 @@ door_hanger_button2=OK!
 infobar_screenshare_no_guest_message=As soon as your friend joins, they will be able to see any tab you click on.
 infobar_screenshare_browser_message2=You are sharing your tabs. Any tab you click on can be seen by your friends
 infobar_screenshare_browser_message3=You are now sharing your tabs. Your friend will see any tab you click on.
-infobar_screenshare_stop_sharing_message=You are no longer sharing your tabs
+infobar_screenshare_stop_sharing_message2=You are no longer sharing your tabs.
+infobar_screenshare_stop_no_guest_message=You have stopped sharing your tabs. When your friend joins, they won't be able to see anything until you restart sharing.
```
dmose: This uses a regular single quote, rather than a unicode single quote. It came up in a recent thread that this will now turn tests orange on mozilla-central.  I've added a commit that fixes this.

dmose:
OK, so I couldn't actually push my commits to this branch, @chrafuse needs to give me explicit access permission, I think.   My commits are here: 

https://github.com/mozilla/loop/commit/477eb614a44c49259987f4d5c7d4b0e8e9542e79
https://github.com/mozilla/loop/commit/309670c0688cfb05da7622f794a61bc1e008d8fb

dmose:
r=dmose, with those commits added

PR #245, opened 2016-03-04
Bug 1253559 - Adjust Roomlist Item dropdown menu so it does not get cut off

daveccrit:
r? @dmose, @chrafuse, @Mardak


daveccrit:
Adding some unit testing, not quite ready yet

chrafuse:
Looks good. Ship it! r=chrafuse

PR #247, opened 2016-03-05
Bug 1252306 - Link clicker intro screen code upgrade

chrafuse:
@dmose Now sets state on close or ok, got it! buttons from one callback method.


daveccrit:
Looks good. Ship it

PR #255, opened 2016-03-08
Bug 1250444 - Local speech bubbles should be grey, remote's should be…

mancas:
… blue


chrafuse:
Looks good. r=chrafuse

PR #258, opened 2016-03-08
Bug 1254511 - Improve functional tests so that they can work against production and pre-release production builds for QA.

Standard8:


daveccrit:
Ship it! +r from me

PR #265, opened 2016-03-09
Bug 1254945 - If text chat consists of entirely context tiles, then it doesn't automatically scroll.

Standard8:


shared/test/loop_mocha_utils.js
```diff
@@ -270,7 +270,7 @@ var LoopMochaUtils = (function(global, _) {
     var consoleWarn = console.warn;
     var consoleError = console.error;
     console.warn = function() {
-      var args = Array.slice(arguments);
+      var args = Array.prototype.slice.call(arguments);
```
Standard8, line 273: These changes were to fix the failure reporting in Karma when it was running the tests in Google Chrome - Chrome requires the more explicit form.

PR #267, opened 2016-03-09
Bug 1254940 - If the link generator exits a room whilst sharing is pa…

mancas:
…used, and then re-enters, the link-clicker display is messed up


daveccrit:
looks good, ship it

PR #270, opened 2016-03-10
Bug 1255410 - Ensure leaving the room works properly when using the leave button on the popped out conversation window.

Standard8:


dmose:
I noticed this in the console after clicking leave with the patch:

16:45:20.719 Loop:Unable to send event through to target: Message port has been destroyed1 MozLoopAPI.jsm:1267

I'm assuming this is actually OK, but I don't have a complete model of the flow in my head here.  Can you confirm?

add-on/panels/js/roomViews.jsx
```diff
@@ -412,8 +412,14 @@ loop.roomViews = (function(mozL10n) {
      */
     leaveRoom: function() {
       if (this.state.used) {
+        // The room has be used, so we might want to display the feedback view.
```
dmose: s/has be/has been/

add-on/panels/js/roomViews.jsx
```diff
@@ -412,8 +412,14 @@ loop.roomViews = (function(mozL10n) {
...
+        // Therefore we leave the room to ensure that we have stopped sharing etc,
+        // then trigger the call terminated handler that'll do the right thing
+        // for the feedback view.
         this.props.dispatcher.dispatch(new sharedActions.LeaveRoom());
+        this.props.onCallTerminated();
```
dmose, line 420: This is certainly an improvement from what was there before.  

That said, it feels odd that the closeWindow is hidden behind the onCallTerminated in one of the if clauses but not the other.

This combined with the fact that closeWindow implicitly causes an unload event to be fired, which then is handle and causes a WindowClose action to be dispatched, makes our window shutdown code fairly convoluted and hard to follow without a bunch of work.  I'm not sure if there's anything to be done about that here or not, though.  Maybe onCallTerminated (and handleCallTerminated that implements it) should be named in a way to describe what it does, rather than when it executes?

add-on/panels/test/roomViews_test.js
```diff
@@ -452,6 +452,40 @@ describe("loop.roomViews", function() {
         sinon.match.hasOwn("name", "setMute"));
     });
 
+    describe("#leaveRoom", function() {
+      it("should close the window when leaving a room that hasn't been used", function() {
+        view = mountTestComponent();
+
```
dmose: No need for a blank line here; both the above code line and the below code line are part of the setup stanza and should live together.  Same for the other tests here.

dmose:
r=@dmose with any comments appropriately addressed.

add-on/panels/js/roomViews.jsx
Standard8, line 420: At this stage, I'd rather keep it simple - especially as I want to quickly uplift this. I'm not entirely happy with the whole flow here - it feels like the feedback view display should be controlled with the `activeRoomStore`, and in the `DesktopRoomConversationView` rather than in the conversation view. However, we got here because we had two call types, and I can't decide at this stage if we should simplify it more or not.

Standard8:
> 16:45:20.719 Loop:Unable to send event through to target: Message port has been destroyed1 MozLoopAPI.jsm:1267
> 
> I'm assuming this is actually OK, but I don't have a complete model of the flow in my head here. Can you confirm?

I think that's just a message attempting to be returned to the window, after its been unloaded. I think I saw it once yesterday, but I can't reproduce it today unfortunately. So I think we'll push through and keep an eye out for any follow-ups.

PR #272, opened 2016-03-11
Bug 1255440 - Local and remote streams needs to be moved to the top o…

mancas:
…f the text chat bar


chrafuse:
Manually tested and came across a related but separate bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1258923) Looks good!  Ship it!  r=chrafuse

PR #273, opened 2016-03-11
Bug 1255750 - If the link clicker join a room whilst sharing is pause…

mancas:
…d, the display is messed up


chrafuse:
Tested locally manually and unit tests.  I cannot run mochi atm, so you may want to "make test" before landing. Looks good. Ship it!  r=chrafuse

PR #274, opened 2016-03-11
Bug 1239965 - Door hanger prompts the user to share on Hello as he copies the URL content

Mardak:


add-on/chrome/bootstrap.js
```diff
@@ -346,6 +348,132 @@ var WindowListener = {
...
+        else if (ticket >= COPY_MAX_TICKET) {
+          this.addCopyPanel();
+          return;
+        }
+
+        // Look up the DNS A-record of a throttler hostname to decide to show.
+        this._dns.asyncResolve(
+          Services.prefs.getCharPref("loop.copy.throttler"),
+          this._dns.RESOLVE_DISABLE_IPV6,
+          (request, record) => {
```
dmose: Please break the callback out into its own named function for improved readability of the parent function.

add-on/chrome/bootstrap.js
```diff
@@ -346,6 +348,132 @@ var WindowListener = {
...
+
+          // Open up the copy panel at the loop button.
+          LoopUI.PanelFrame.showPopup(
+            window,
+            LoopUI.toolbarButton.anchor,
+            "loop-copy",
+            null, // Origin
+            "chrome://loop/content/panels/copy.html",
+            null, // Size
+            iframe => {
```
dmose: Please also break this callback out into its own named function for improved readability of the parent function.

Mardak: E.g., `let onIframe = iframe => {` next to the `let addTelemetry` above?

add-on/chrome/bootstrap.js
Mardak: Moved to `let onDNS` with comments. Also collapsed the multiline-argument-style asyncResolve call to singleline-argument-style.

add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -69,6 +69,20 @@ const SHARING_SCREEN = {
   RESUMED: 1
 };
 
+/**
+ * Values that we segment copy panel action telemetry probes into.
+ *
+ * @type {{SHOWN: Number, NO_AGAIN: Number, NO_NEVER: Number, YES_AGAIN: Number,
+ *         YES_NEVER: Number}}
+ */
```
dmose, line 76: These all want detailed documentation, I think.  Perhaps an @enum a la http://usejsdoc.org/tags-enum.html

dmose:
I found the throttling code to be difficult to grok without reading through detailed comments in the bug.  I think this is actually partly do to a related issue: LoopUI is (and has been, before this patch) in danger of falling down the same rathole that MozLoopService, and, to some degree, MozLoopAPI, have fallen down: no separation of concerns.  Which is to say that we're not adhering to the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle). 

However, I think there's a pretty easy opportunity here: if we split out the throttler code into its own object, where the DNS algorithm is documented in detail and tested separately from the rest of the code, it'll be much easier to grok.  I suspect it wants to be in its own JSM, though if that's truly painful, I'm open to other possibilities.

I actually think LoopUI probably wants to be broken up even further, but we don't need to entrain that work here.  Just leading the way by making this code a clean example for future modifications would be a real win, I think.

Happy to chat about this in more detail and/or verbally as well.

Mardak:
Split out Throttler to a separate object in bootstrap.js with a separate mochitest. Added additional button/checkbox functionality with additional tests. All mochitests passing on a test export.

add-on/chrome/bootstrap.js
```diff
@@ -352,6 +353,105 @@ var WindowListener = {
         }
       },
 
+      /**
+       * Maybe add the copy panel if it's not throttled and passes other checks.
+       * @return {Promise} Resolved when a decision and panel adding are done.
```
dmose: s/panel adding/and possibly panel-adding/ would be a bit clearer.

add-on/panels/js/copy.jsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+  });
+
+  /**
+   * Copy panel initialization.
+   */
+  function init() {
+    let { LoopAPI } = Cu.import("chrome://loop/content/modules/MozLoopAPI.jsm", {});
+    let requests = ["GetAllStrings", "GetLocale", "GetPluralRule"];
+    return Promise.all(requests.map(name => new Promise(resolve => {
```
dmose, line 78: Please do something to improve the skimmability of this, at the very least adding a comment.

dmose, line 78: Splitting out a getFromLoopAPI function or something like that could be another possibility.

add-on/chrome/test/mochitest/browser_copypanel.js
```diff
@@ -0,0 +1,114 @@
...
+
+// Even with max threshold, no panel if already shown.
+add_task(function* test_init_copy_panel_already_shown() {
+  LoopUI.removeCopyPanel();
+  Services.prefs.setIntPref("loop.copy.ticket", -1);
+  Services.prefs.setBoolPref("loop.copy.shown", true);
+
+  yield LoopUI.maybeAddCopyPanel();
+
+  Assert.equal(document.getElementById("loop-copy-notification-panel"), null, "copy panel doesn't exist already shown");
```
dmose: The failure text wants a tweak, I think.

add-on/chrome/bootstrap.js
```diff
@@ -352,6 +353,105 @@ var WindowListener = {
...
+          });
+        };
+
+        // Override the default behavior of the copy command.
+        let controller = gURLBar._copyCutController;
+        controller._doCommand = controller.doCommand;
+        controller.doCommand = function() {
+          // Do the normal behavior first.
+          this._doCommand.apply(this, arguments);
+
```
dmose, line 440: Let's use "controller" instead of "this" for clarity.

add-on/chrome/bootstrap.js
```diff
@@ -926,6 +1026,7 @@ var WindowListener = {
     XPCOMUtils.defineLazyModuleGetter(LoopUI, "PlacesUtils", "resource://gre/modules/PlacesUtils.jsm");
 
     LoopUI.init();
+    window.LoopThrottler = Throttler;
     window.LoopUI = LoopUI;
```
dmose, line 1049: If LoopThrottler is only exported for testing, please be explicit about that here.

add-on/chrome/test/mochitest/browser_copypanel.js
```diff
@@ -0,0 +1,114 @@
...
+  LoopUI.addCopyPanel();
+  gURLBar.focus();
+  gURLBar.select();
+  goDoCommand("cmd_copy");
+
+  let iframe = document.getElementById("loop-copy-panel-iframe");
+  return new Promise(resolve => {
+    iframe.addEventListener("DOMContentLoaded", () => {
+      iframe.parentNode.addEventListener("popuphidden", () => resolve(histogram));
+      setTimeout(() => onIframe(iframe));
```
dmose, line 64: Let's add a comment explaining why calling without an interval is necessary here.

add-on/chrome/test/mochitest/browser_copypanel.js
```diff
@@ -0,0 +1,114 @@
...
+
+  Assert.equal(win.document.getElementById("loop-copy-notification-panel"), null, "copy panel doesn't exist for private browsing");
+  Assert.equal(Services.prefs.getIntPref("loop.copy.ticket"), -1, "ticket should be unchanged");
+
+  win.close();
+});
+
+// Helper function for testing clicks on the copy panel.
+function testClick(domain, onIframe) {
+  let histogram = Services.telemetry.getHistogramById("LOOP_COPY_PANEL_ACTIONS");
```
dmose, line 49: Adding a comment explaining why domain is here (for test debuggability) would be great.

add-on/chrome/bootstrap.js
```diff
@@ -968,6 +1070,75 @@ var WindowListener = {
...
+      else if (ticket >= this.TICKET_LIMIT) {
+        resolve();
+        return;
+      }
+
+      // Handle responses from the DNS resolution service request.
+      let onDNS = (request, record) => {
+        // Use a specific part of the A-record IP address depending on the
+        // channel. I.e., 127.[release/beta/other].[aurora].[nightly].
+        let index = 1;
```
dmose, line 1134: It sounds like we want the split to be different, with release & beta each getting their own channel, IIRC.

add-on/chrome/bootstrap.js
```diff
@@ -968,6 +1070,75 @@ var WindowListener = {
...
+        Services.prefs.setIntPref(prefTicket, ticket);
+      }
+      // Short circuit if the special ticket value indicates we're good to go.
+      else if (ticket >= this.TICKET_LIMIT) {
+        resolve();
+        return;
+      }
+
+      // Handle responses from the DNS resolution service request.
+      let onDNS = (request, record) => {
```
dmose, line 1131: Since we've got a much smaller object here, I'd propose that onDNS move to a top-level function, rather than something lexically scoped, for readability.

add-on/chrome/bootstrap.js
```diff
@@ -968,6 +1070,75 @@ var WindowListener = {
...
+          case "aurora":
+            index = 2;
+            break;
+          case "nightly":
+            index = 3;
+            break;
+        }
+
+        // Check if the 8-bit threshold (0-255) exceeds the ticket (0-254).
+        let threshold = record && record.getNextAddrAsString().split(".")[index];
```
dmose, line 1147: While exceedingly elegant, this line has more than one layer of abstractions, and depends on knowing that IPv4 addresses when passed to split this way will create the array you want.  Please at least split this up with temporaries for readability.

add-on/chrome/test/mochitest/browser_copypanel.js
```diff
@@ -0,0 +1,114 @@
...
+// Click the cancel button with checkbox.
+add_task(function* test_click_no_never() {
+  let histogram = yield testClick("no.never", iframe => {
+    iframe.contentDocument.querySelector(".copy-toggle-label").click();
+    iframe.contentDocument.querySelector(".copy-button").click();
+  });
+
+  Assert.equal(document.getElementById("loop-copy-notification-panel"), null, "copy panel removed");
+  Assert.strictEqual(histogram.snapshot().counts[gConstants.COPY_PANEL.NO_NEVER], 1, "triggered telemetry count for no never");
+});
```
dmose, line 119: There's nothing here that seems to test that the semantics of the _NEVER _AGAIN works.   It seems like that shouldn't be too easy, though possibly even easier to do in a unit-style test rather than by driving the UI in an integration test.  Either strategy could be added to this file, I'd say.

add-on/chrome/test/mochitest/browser_throttler.js
```diff
@@ -0,0 +1,167 @@
+/* Any copyright is dedicated to the Public Domain.
+   http://creativecommons.org/publicdomain/zero/1.0/ */
+
```
dmose, line 3: For bonus points, port this to xpcshell so that it runs faster.  I don't think we want to block on that if it's non-trivial, though, so at least put an XXX comment suggesting that in here.

dmose, line 3: These unit tests are nicely done; good work!

add-on/panels/css/copy.css
```diff
@@ -0,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
dmose, line 1: Nice clean CSS too!

add-on/panels/js/copy.jsx
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Copy panel view.
+   */
+  let CopyView = React.createClass({
+    getInitialState() {
+      return { checked: false };
+    },
+
+    _dispatch(accept) {
+      window.dispatchEvent(new CustomEvent("CopyPanelClick", {
```
dmose, line 27: A comment describing what this event is intended to cause the chrome code to do would be nice.

add-on/panels/js/copy.jsx
```diff
@@ -0,0 +1,99 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global Components */
+let { utils: Cu } = Components;
+
```
dmose, line 7: If you could try setting up some simple unit tests here stubbing out the chrome things, that'd be great.  That said, I'd suggest timeboxing this so that we can still try and land tomorrow.

add-on/panels/js/copy.jsx
```diff
@@ -0,0 +1,99 @@
...
+var loop = loop || {};
+loop.copy = (mozL10n => {
+  "use strict";
+
+  /**
+   * Copy panel view.
+   */
+  let CopyView = React.createClass({
```
dmose, line 15: You should be able to use the [PureRenderMixin](https://facebook.github.io/react/docs/pure-render-mixin.html) here for a potential slight perf boost.

dmose:
r=dmose.  That said, if you have any code you'd like me to review before landing, I'm happy to do that.

add-on/panels/js/copy.jsx
Mardak, line 7: Stubbed as fake-components now used for desktop local tests, karma, and ui showcase.

add-on/chrome/test/mochitest/browser_throttler.js
Mardak, line 3: Throttler is created in bootstrap.js and not a service, so it would be tricky to test via xpcshell.

add-on/chrome/test/mochitest/browser_copypanel.js
Mardak, line 119: Added a test that tries to show the copy panel again and makes sure it doesn't.

add-on/chrome/bootstrap.js
Mardak, line 1131: onDNS makes use of a couple variables in scope: `ticket` and `prefPrefix` so moving onDNS out would require some passing around those variables to a makeOnDNS function to return an onDNS callback.

add-on/chrome/bootstrap.js
Mardak, line 1134: Yup. Fixed.

PR #278, opened 2016-03-14
Bug 1255491 - Disable pop-out feature in conversation window

mancas:


daveccrit:
Looks good. r=dcritch

Standard8:
Per discussion in standup today, please ship this with `loop.conversationPopOut.enabled` as `true` - there's other code landing soon that should fix the e10s issues this works around, but we want this just in case we need to turn things off.

PR #283, opened 2016-03-15
Bug 1256694 - Use a better maximum version definition for the add-on.

Standard8:


Makefile
```diff
@@ -12,6 +12,8 @@ LOOP_PRIVACY_WEBSITE_URL := $(shell echo $${LOOP_PRIVACY_WEBSITE_URL-"https://ww
 LOOP_LEGAL_WEBSITE_URL := $(shell echo $${LOOP_LEGAL_WEBSITE_URL-"https://www.mozilla.org/about/legal/terms/firefox-hello/"})
 LOOP_PRODUCT_HOMEPAGE_URL := $(shell echo $${LOOP_PRODUCT_HOMEPAGE_URL-"https://www.firefox.com/hello/"})
 FIREFOX_VERSION=48.0
+# Taken from http://hg.mozilla.org/mozilla-central/file/5e14887312d4//old-configure.in#l8032
+MOZ_APP_MAXVERSION=`echo $(FIREFOX_VERSION) | sed "s|\(^[0-9]*\).*|\1|"`.*
```
Mardak: Are we planning on using `FIREFOX_VERSION` for something? It's passed in to the preprocessor as well but not used in `install.rdf.in`. So perhaps just have a `FIREFOX_MAXVERSION=48.*` to not need the `sed`/comment.

Standard8: We need to keep it as MOZ_APP_MAXVERSION (due to mozilla-central), but you're right, there's no point in doing the sed. I've changed that, I've also changed install.rdf into its own target, so that we're not unnecessarily firing up the preprocessor when we don't need to.

Makefile
```diff
@@ -179,6 +182,13 @@ $(DIST)/add-on/chrome/content/preferences/prefs.js: $(VENV) add-on/preferences/p
 	  python $(VENV)/lib/python2.7/site-packages/mozbuild/preprocessor.py \
 	         -D LOOP_BETA=1 -o $@ add-on/preferences/prefs.js
 
+$(BUILT)/add-on/install.rdf: $(VENV) Makefile add-on/install.rdf.in
+	@mkdir -p $(@D)
+	. $(VENV)/bin/activate && \
+	  python $(VENV)/lib/python2.7/site-packages/mozbuild/preprocessor.py \
+	         -D MOZ_APP_MAXVERSION=$(MOZ_APP_MAXVERSION) \
+	         -o $@ add-on/install.rdf.in
```
Mardak: Perhaps this would be too much Makefile magic.. `-o $@ $<` and switching the order of prerequisites: `add-on/install.rdf.in $(VENV) Makefile` ;)

(The current style matches the existing style of other rules.)

Standard8: I decided to go for it. I think its reasonable either way, but $< cuts the repetition.

PR #288, opened 2016-03-15
Bug 1257009 - The link generator shouldn't try to setup tab browsing every time a link clicker leaves the room.

Standard8:


add-on/panels/js/roomViews.jsx
```diff
@@ -401,7 +401,10 @@ loop.roomViews = (function(mozL10n) {
       }
 
       // Automatically start sharing a tab now we're ready to share.
+      // Only do this if we're not already connected, e.g. not in the case the
+      // remote person has just left.
       if (this.state.roomState !== ROOM_STATES.SESSION_CONNECTED &&
+          this.state.roomState !== ROOM_STATES.HAS_PARTICIPANTS &&
           nextState.roomState === ROOM_STATES.SESSION_CONNECTED) {
         this.props.dispatcher.dispatch(new sharedActions.StartBrowserShare());
```
Mardak, line 409: There's a private `isConnectedToRoom(state)` helper function https://github.com/mozilla/loop/blob/master/shared/js/mixins.js#L482 that might help make this more readable by grouping things together. No need for the separate function, but maybe more of..

```
// Now that we're ready to share, automatically start sharing a tab only if we're not already connected to the room, e.g., ..
if (nextState.roomState === ROOM_STATES.SESSION_CONNECTED &&
    !(this.state.roomState === ROOM_STATES.SESSION_CONNECTED || this.state.roomState === ROOM_STATES.HAS_PARTICIPANTS))
```

In particular, 1) the sentence ordering of the comment matches up with the condition and 2) clarifying "**connected** to the room" (as that's different from SESSION_**CONNECTED**).

Or perhaps have a local variable `connectedToRoom` and check `!connectedToRoom`.

PR #293, opened 2016-03-17
Bug 1256357 - Enable Hello in e10s windows by default. r=Standard8

mikedeboer:


Standard8:
This will be in tomorrow's nightly, but landing now so that its on the release branch & ready.

PR #294, opened 2016-03-17
Bug 1256362 - Remove the 'Start a conversation' menuitem when in private browsing mode.

Standard8:


add-on/chrome/bootstrap.js
```diff
@@ -299,7 +299,9 @@ var WindowListener = {
           }
         });
 
-        this.addMenuItem();
+        if (!PrivateBrowsingUtils.isWindowPrivate(window)) {
+          this.addMenuItem();
+        }
```
mancas, line 306: Shouldn't we add a return statement here? If we're in a private window and the icon is hide, we shouldn't add observers and the other stuff, right?

Standard8, line 306: Good catch, you're totally right we don't need to add those observers (and it stops the tooltip getting changed when it shouldn't be).

PR #296, opened 2016-03-18
Bug 1257154 - Switch to getting DOM title updates via a frame script, and make the add-on truely multiprocess compatible to avoid shims.

Standard8:


add-on/chrome/bootstrap.js
```diff
@@ -550,7 +569,11 @@ var WindowListener = {
 
         this._hideBrowserSharingInfoBar();
         gBrowser.tabContainer.removeEventListener("TabSelect", this);
-        gBrowser.removeEventListener("DOMTitleChanged", this);
+
+        if (titleChangedListener) {
+          this.mm.removeMessageListener("loop@mozilla.org:DOMTitleChanged",
+            titleChangedListener);
```
mikedeboer: Please do `titleChangedListener = null;`, otherwise this whole exercise is for naught...

add-on/chrome/bootstrap.js
```diff
@@ -313,6 +321,13 @@ var WindowListener = {
           return;
         }
 
+        // Load the frame script into any tabs, plus any that get created.
```
mikedeboer: nit: s/tabs/tab/

mikedeboer: nit: 'plus any tab that get created in the future.'

add-on/chrome/bootstrap.js
```diff
@@ -313,6 +321,13 @@ var WindowListener = {
...
+        // Note: We'd ideally do this only when tab sharing, but as we can't
```
mikedeboer: I think it's unnecessary to put this comment here... the way loading frame scripts work is not going to change in the future, so this sounds a bit like reminiscing :wink:

add-on/chrome/bootstrap.js
```diff
@@ -522,9 +537,13 @@ var WindowListener = {
           gBrowser.tabContainer.addEventListener("TabSelect", this);
           this._listeningToTabSelect = true;
 
+          titleChangedListener = this.handleDOMTitleChanged.bind(this);
+
           // Watch for title changes as opposed to location changes as more
           // metadata about the page is available when this event fires.
-          gBrowser.addEventListener("DOMTitleChanged", this);
+          this.mm.addMessageListener("loop@mozilla.org:DOMTitleChanged",
+            titleChangedListener);
```
mikedeboer, line 542: You 'could' put the assignment here as well: `titleChangedListener = this.handleDOMTitleChanged.bind(this));`... up to you.

add-on/chrome/modules/tabFrame.js
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+/* global sendAsyncMessage */
```
mikedeboer, line 7: Don't you need to add `addEventListener` to this statement here? If it's already part of .eslintrc, you might want to put this one in there as well.

add-on/chrome/modules/tabFrame.js
```diff
@@ -0,0 +1,22 @@
...
+
+// Listen for when the title is changed and send a message back to the chrome
+// process.
+addEventListener("DOMTitleChanged", function(event) {
+  sendAsyncMessage("loop@mozilla.org:DOMTitleChanged", {
+    details: "titleChanged"
+  }, {
+     target: event.target
+  });
+}, false);
```
mikedeboer: The `false` is default, you may omit it.

mikedeboer: I usually put simple, short object literals on a single-line, but that's a matter of taste :wink:

Mardak: If you want to use new fancy es6... ;)

```
addEventListener("DOMTitleChanged", ({ target }) => {
  sendAsyncMessage("loop@mozilla.org:DOMTitleChanged", { details: "titleChanged" }, { target });
});
```

(Destructured assignment, arrow function, shorthand property)

add-on/chrome/modules/tabFrame.js
Standard8, line 7: `addEventListener` is something in the eslint built-in configuration (I think as a result of telling it the code is loaded into a browser), we don't specify it as a global anywhere - I didn't really want to add it very globally, to avoid possible leaks/incorrect usage elsewhere.

mikedeboer:
r=me!

PR #299, opened 2016-03-18
Bug 1253013

daveccrit:
r? @Standard8, @dmose


dmose:
I'll do the review.  I've just proposed a different architecture in the bug.

PR #301, opened 2016-03-21
Bug 1258329 - Allow easy generation of a "dev" xpi for uploading to AMO's dev channel.

Standard8:


Makefile
```diff
@@ -57,6 +57,14 @@ BABEL := $(NODE_LOCAL_BIN)/babel
 ESLINT := $(NODE_LOCAL_BIN)/eslint
 FLAKE8 := $(NODE_LOCAL_BIN)/flake8
 
+ifdef DEV_XPI
+LOOP_XPI_DATE=`date +"%Y%m%d%H%M"`
```
dmose, line 63: Put hyphens between each field for improved human readability?

dmose, line 63: Assuming AMO allows this, of course...

dmose, line 63: Maybe even a hyphen at the beginning, as a separator from the package name.

Makefile
```diff
@@ -170,23 +178,26 @@ vendor_libs: $(BACKBONE_OBJS) $(CLASSNAME_OBJS) $(LODASH_OBJS) $(REACT_OBJS) \
              $(BUILT)/test/vendor/mocha.js $(BUILT)/test/vendor/mocha.css \
              $(BUILT)/test/vendor/chai.js $(BUILT)/test/vendor/chai-as-promised.js
 
-$(BUILT)/add-on/chrome/content/preferences/prefs.js: $(VENV) add-on/preferences/prefs.js
+$(BUILT)/add-on/chrome/content/preferences/prefs.js: add-on/preferences/prefs.js $(VENV) Makefile
```
dmose: Suggest using \ to split each dependency out to its own line for readability, here, and in the other instances of this change.

Makefile
```diff
@@ -57,6 +57,14 @@ BABEL := $(NODE_LOCAL_BIN)/babel
 ESLINT := $(NODE_LOCAL_BIN)/eslint
 FLAKE8 := $(NODE_LOCAL_BIN)/flake8
 
+ifdef DEV_XPI
+LOOP_XPI_DATE=`date +"%Y%m%d%H%M"`
+LOOP_DEV_XPI=-D LOOP_DEV_XPI=1
```
dmose: I'd suggest calling this LOOP_DEV_XPI_DEFS or something.  Right now, reading the call sites is a bit confusing, as it looks like there might be an XPI file in the middle of the commands.

Makefile
```diff
@@ -170,23 +178,26 @@ vendor_libs: $(BACKBONE_OBJS) $(CLASSNAME_OBJS) $(LODASH_OBJS) $(REACT_OBJS) \
...
+$(BUILT)/add-on/chrome/content/preferences/prefs.js: add-on/preferences/prefs.js $(VENV) Makefile
 	mkdir -p $(@D)
 	. $(VENV)/bin/activate && \
 	  python $(VENV)/lib/python2.7/site-packages/mozbuild/preprocessor.py \
-	         -D DEBUG=1 -D LOOP_BETA=1 -o $@ add-on/preferences/prefs.js
+	         -D DEBUG=1 -D LOOP_BETA=1 $(LOOP_DEV_XPI) \
+	         -o $@ $<
 
-$(DIST)/add-on/chrome/content/preferences/prefs.js: $(VENV) add-on/preferences/prefs.js
+$(DIST)/add-on/chrome/content/preferences/prefs.js: add-on/preferences/prefs.js $(VENV) Makefile
```
dmose: So I'm wondering if this  change wants to be made to the add-on_dist target, because doing a "make dist_xpi" doesn't produce a development XPI.  Or is there some piece of the machinery that has yet to be landed?

Makefile
Standard8, line 63: The number rules are complex, but I think that dashes aren't allowed. I don't think we really need them here either tbh.

Makefile
Standard8: Ok, not strictly part of this bug, but I've adjusted deps around so that dist_xpi now works from a clean build.

Makefile
```diff
@@ -57,6 +57,14 @@ BABEL := $(NODE_LOCAL_BIN)/babel
 ESLINT := $(NODE_LOCAL_BIN)/eslint
 FLAKE8 := $(NODE_LOCAL_BIN)/flake8
 
+ifdef DEV_XPI
```
dmose, line 62: Maybe add a comment about how this is expected to be set?  (eg the command-line)

Makefile
```diff
@@ -387,9 +406,11 @@ dist_export:
 	@mkdir -p $(DIST_EXPORT_DIR)
 	# Do a basic copy of the add-on.
 	$(RSYNC) $(BUILT)/$(ADD-ON)/* $(DIST_EXPORT_DIR)
-	# Use the install.rdf.in rather than install.rdf.
+	# Use a modified install.rdf.in rather than install.rdf.
 	rm -f $(DIST_EXPORT_DIR)/install.rdf
-	$(RSYNC) $(ADD-ON)/install.rdf.in $(DIST_EXPORT_DIR)
+	# mozilla-* doesn't need the date field.
```
dmose: mozilla-\* is a fairly "insider" term.  Maybe "mozilla-central and other firefox repositories"?

Makefile
```diff
@@ -466,7 +487,7 @@ endif
 #
 
 .PHONY: build
-build: add-on standalone ui vendor_libs
+build: add-on standalone ui
```
dmose, line 492: Good idea to push the vendor_libs dep down.

add-on/preferences/prefs.js
```diff
@@ -1,7 +1,12 @@
 pref("loop.enabled", true);
 pref("loop.remote.autostart", true);
+#ifdef LOOP_DEV_XPI
```
dmose, line 3: Might be simpler to just use DEV_XPI here too, to keep the define names consistent across various places.  Up to you.

dmose:
r=dmose, with any appropriate changes.

add-on/preferences/prefs.js
Standard8, line 3: I wanted to use LOOP_DEV_XPI so that when it gets into mozilla-central, its clear its a variable belonging to us.

PR #303, opened 2016-03-22
Bug 1254520 - Opening your own room link should have a functional test

mancas:


test/functional/manifest.ini
```diff
@@ -3,4 +3,4 @@ b2g = false
 browser = true
 qemu = false
 
-[test_1_browser_call.py]
+[test_2_linkclicker.py]
```
mancas, line 7: This change is just for testing purpose

test/functional/helpers.py
```diff
@@ -0,0 +1,142 @@
...
+import pyperclip
+import re
+import urlparse
+import time
+
+from serversetup import ROOMS_WEB_APP_URL_BASE
+from config import USE_LOCAL_STANDALONE
+
+
+def wait_for_element_displayed(self, by, locator, timeout=None):
```
Standard8: I think what might be slightly nicer, is to change the functions in this file to be within a class. The class. Something like:

``` python
class LoopTestDriver():
    def setUp(self, marionette):
        self.marionette = marionette
```

Then in `test_*.py` we should be able to do:

``` python
class Test1BrowserCall(LoopTestDriver, MarionetteTestCase):
    def setUp():
        # start server
        self.loop_test_servers = LoopTestServers()

        MarionetteTestCase.setUp(self)
        LoopTestDriver.setUp(self, marionette)

        ...
```

I think that will work, if not https://github.com/mozilla/talkilla/tree/master/test/functional may also help. It is not quite the same - as we don't have web driver, but the ideas should be similar.

Standard8: Oh, and this also means that we should be able to maintain the existing:

``` python
self.switch_to_panel()
```

style of calling the functions.

test/functional/test_2_linkclicker.py
```diff
@@ -0,0 +1,119 @@
...
+        self.marionette.restart(in_app=True)
+
+        # this is browser chrome, kids, not the content window just yet
+        self.marionette.set_context("chrome")
+
+    def navigate_to_standalone(self, url):
+        self.switch_to_standalone()
+        self.marionette.navigate(url)
+
+    def standalone_check_view(self):
```
Standard8: I think this would be clearer if it was named as `standalone_check_own_link_view`

test/functional/test_2_linkclicker.py
```diff
@@ -0,0 +1,119 @@
...
+        self.local_leave_room()
+
+        # load the link clicker interface into the current content browser
+        self.navigate_to_standalone(room_url)
+
+        self.standalone_check_view()
+
+        self.standalone_join_own_room()
+
+        self.load_homepage()
```
Standard8, line 109: Maybe just add a comment here `# Ensure we're on a different page so that we can navigate back to the room url later.`

PR #306, opened 2016-03-22
Bug 1258834 - Disable Loop's e10s for FF 46 and Conversation pop-outs for FF 47.

Standard8:


dmose:
rs=dmose

add-on/chrome/bootstrap.js
```diff
@@ -1039,6 +1039,14 @@ function loadDefaultPrefs() {
       }
     }
   });
+
+  if (Services.vc.compare(Services.appinfo.version, "47.0a1") < 0) {
+    branch.setBoolPref("loop.remote.autostart", false);
```
mikedeboer, line 1044: We don't need to do this, because we also check if the current window is in e10s mode - in other words: when e10s is disabled in the browser, we will not boot conversation windows in e10s mode.
We follow the rules of the host.

Standard8, line 1044: I believe e10s is currently enabled for a few users on 46. I think it gets turned off around b6 time, though I'm not sure (b6 will be after we ship this).

As I'm not sure, I don't want us turning it on by accident and causing lots of crashes/other issues ;-)

mikedeboer, line 1044: True, via the experiment. But that's only for the beta audience. e10s will _not_ be turned on in 46 GA.

Standard8, line 1044: Right, but I don't want to mess up the test audience during any possible overlap. Nor do I want testers trying to use it accidentally ;-)

Turning the pref off in GA won't actually affect anyone.

mikedeboer, line 1044: True!

PR #307, opened 2016-03-22
Bug 1257947 - fall back to Linux skin for non Win/Mac

rhelmer:


mozilla-autolander-deprecated:

PR #319, opened 2016-03-29
No bug. Update python packages to the latest version and improve handling of our pip requirements.

Standard8:
This does a couple of things:
- Drops us trying to install virtualenv globally when it really should be done by the user (as it might require sudo or whatever) - but adds documentation for the dependency.
- Lists all the dependencies of the python packages, not just the ones we require.
- Update all packages to the latest
  - This also includes updating marionette for which it supplies a couple of new properties that would be useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,6 @@ RSYNC := rsync --archive --exclude='*.jsx'
 
 .PHONY: install
 install: node_modules
-	pip install -r require.pip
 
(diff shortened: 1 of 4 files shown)


mozilla-autolander-deprecated:

daveccrit:
r=dcritch

mozilla-autolander-deprecated:

PR #326, opened 2016-03-31
No bug. Attempt to fix build issues with mocha.* not always being found.

Standard8:
This attempts to fix the issue where sometimes one of the mocha file sources isn't found. By splitting these rules, I think it'll fix the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,10 @@ $(BUILT)/test/vendor:
 $(BUILT)/test/vendor/sinon.js: node_modules/sinon/pkg/sinon.js
 	$(RSYNC) $< $(BUILT)/test/vendor
 
-$(BUILT)/test/vendor/mocha.%: node_modules/mocha/mocha.%
+$(BUILT)/test/vendor/mocha.js: node_modules/mocha/mocha.js
+	$(RSYNC) $< $(BUILT)/test/vendor
+
+$(BUILT)/test/vendor/mocha.css: node_modules/mocha/mocha.css
 	$(RSYNC) $< $(BUILT)/test/vendor
 
 $(BUILT)/test/vendor/chai.js: node_modules/chai/chai.js


mozilla-autolander-deprecated:

Makefile
```diff
@@ -163,7 +163,10 @@ $(BUILT)/test/vendor:
 $(BUILT)/test/vendor/sinon.js: node_modules/sinon/pkg/sinon.js
 	$(RSYNC) $< $(BUILT)/test/vendor
 
-$(BUILT)/test/vendor/mocha.%: node_modules/mocha/mocha.%
+$(BUILT)/test/vendor/mocha.js: node_modules/mocha/mocha.js
+	$(RSYNC) $< $(BUILT)/test/vendor
+
+$(BUILT)/test/vendor/mocha.css: node_modules/mocha/mocha.css
 	$(RSYNC) $< $(BUILT)/test/vendor
```
dmose, line 170: Ha, yes, that totally makes sense.  If it's not installed already, the original rule wouldn't trigger.

r=dmose

dmose, line 170: I believe, though, we can avoid the duplication by only listing the recipe once, while duplicating the targets.  Eg:

``` make
target1:
target2:
  $(RSYNC) ...
```

Standard8, line 170: I just tried it listing the recipe once locally and it doesn't work. I couldn't find anything in make's documentation about it either.

mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

Makefile
dmose, line 170: OK, I must have been confused.  Never mind!

Mardak, line 170: I believe the syntax you're looking for:

``` make
target1: dep1
target2: dep2
target1 target2:
  $(RSYNC) ...
```

PR #327, opened 2016-04-01
Bug 1261332 - Add Autodeploy scripts for the standalone dev server and dev xpis.

Standard8:


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

bin/autodeploy_devxpi.sh
```diff
@@ -0,0 +1,56 @@
...
+# This is designed to be run from a crontab somewhere.
+
+# WARNING: Running this script will modify your repositories.
+
+# The script expectes the following repository setup:
+#   ${BASEDIR}/loop - A clone of https://github.com/mozilla/loop
+#   ${BASEDIR}/loop-client-l10n - A clone of https://github.com/mozilla/loop-client-l10n
+
+# This script expects JPM_API_KEY and JPM_API_SECRET to be present in the environment.
+# See the upload_xpi command in the Makefile.
```
dmose: I'd suggest changing all of this from a comment to [here document](https://bash.cyberciti.biz/guide/Here_documents) and echoing it out as part of the usage message.  It would also be nice if it describes how your repo is changed, and potentially makes a recommendation for how to use it (eg make a scratch repo, use it there, then delete that repo, or maybe do it in your regular repo and then just do a 'git reset --hard master' afterwards or something)...

bin/autodeploy_devxpi.sh
```diff
@@ -0,0 +1,56 @@
...
+DEV_XPI=1 make build
+DEV_XPI=1 make dist_xpi
+
+# Now upload it.
+
+echo "Uploading. This command will fail with:"
+echo "Your add-on has been submitted for review. It passed validation but..."
+echo "That error is expected (https://github.com/mozilla-jetpack/jpm/issues/520)."
+
+make upload_xpi
```
dmose, line 64: You might want to add a "-n" flag that echoes all the commands without running them, and maybe also default it to not uploading, but adding a "-u" switch to upload automagically...  These would likely be useful both for new folks getting acquainted with the tool, and for the times one wants to verifying the thing that is built.   That said, if you want to these things later rather than now, I'm ok with that.

docs/AutomaticDevDeploys.md
```diff
@@ -0,0 +1,39 @@
...
+l10n files, and package/produce output from the result.
+
+## Dev Server Deploys
+
+The standalone dev server is set up with a cronjob that automatically deploys
+every half hour.
+
+The script it uses is `[bin/autodeploy_standalone.sh](../bin/autodeploy_standalone.sh)`.
+
+The script expects repositories in a particular location - see its header for more
```
dmose: If you switch to here documents, you can just suggest seeing the usage message.

bin/autodeploy_standalone.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
dmose, line 1: All the comments from the first script more-or-less apply here.

PR #344, opened 2016-04-08
Add es6 and react rules that are already passing or have a few errors that are auto-fixable.

Mardak:
rs? @Standard8 

I did enable `"react/jsx-boolean-value": "always"` even though it was under `// These ones we don't want to ever enable` as existing style seems to match it in most cases..

These are the number of lines that were autofixed:

```
   1 jsx-boolean-value
   3 jsx-equals-spacing
  25 jsx-space-before-closing
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -45,6 +45,7 @@
 
     // Eslint built-in rules are documented at <http://eslint.org/docs/rules/>
     "array-bracket-spacing": [2, "never"],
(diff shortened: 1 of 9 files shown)


mozilla-autolander-deprecated:

Standard8:
r=Standard8

mozilla-autolander-deprecated:

PR #346, opened 2016-04-08
Bug 1261849 - Hello conversation opens while Firefox is in never reme…

mancas:
…mber history mode

### DIFF
--- a/add-on/chrome/bootstrap.js
+++ b/add-on/chrome/bootstrap.js
@@ -193,6 +193,10 @@ var WindowListener = {
        * @return {Promise}
        */
       openPanel: function(event) {
+        if (PrivateBrowsingUtils.isWindowPrivate(window)) {
+          return Promise.reject();
+        }
+
         return new Promise((resolve) => {
           let callback = iframe => {
             let mm = iframe.QueryInterface(Ci.nsIFrameLoaderOwner).frameLoader.messageManager;
--- a/add-on/chrome/test/mochitest/browser_panel_privateBrowsing.js
+++ b/add-on/chrome/test/mochitest/browser_panel_privateBrowsing.js
@@ -1,0 +1,37 @@
+/* Any copyright is dedicated to the Public Domain.
+   http://creativecommons.org/publicdomain/zero/1.0/ */
+
+"use strict";
+
+var win;
+
+registerCleanupFunction(function* () {
+  if (win) {
+    yield BrowserTestUtils.closeWindow(win);
+  }
+});
+
+add_task(function* test_panel_should_be_shown() {
+  registerCleanupFunction(function* () {
+    cleanupPanel();
+  });
+
+  yield window.LoopUI.openPanel();
+
+  Assert.notEqual(document.getElementById("loop-notification-panel"), null, "Panel exists");
+});
+
+add_task(function* test_init_copy_panel_private() {
+  let rejected = false;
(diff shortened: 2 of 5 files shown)


add-on/chrome/bootstrap.js
```diff
@@ -211,6 +211,11 @@ var WindowListener = {
             });
           };
 
+          if (PrivateBrowsingUtils.isWindowPrivate(window)) {
+            resolve(null);
+            return;
+          }
```
Mardak: This doesn't need to be inside the `Promise` callback. You can check immediately inside `openPanel` and `return Promise.resolve()` to short circuit.

Mardak: I'm also not sure if we should `reject()` instead as the UITour code will call an optional `aOpenCallback` on success: https://dxr.mozilla.org/mozilla-central/source/browser/components/uitour/UITour.jsm#1720-1742

Mardak: Although, it's quite possible the only uses of the `Promise` is for testing, e.g., https://dxr.mozilla.org/mozilla-central/source/browser/components/uitour/test/head.js#167-174 to indicate the test should continue. We should have a mochitest similar to https://dxr.mozilla.org/mozilla-central/source/browser/components/uitour/test/browser_showMenu_controlCenter.js to make sure the panel opens in non-private browsing and doesn't in private browsing.

add-on/chrome/test/mochitest/browser_panel_privateBrowsing.js
```diff
@@ -0,0 +1,20 @@
...
+
+add_task(function* test_init_copy_panel_private() {
+  let win = OpenBrowserWindow({ private: true });
+  yield new Promise(resolve => win.addEventListener("load", resolve));
+  yield win.LoopUI.openPanel();
+
+  Assert.equal(win.document.getElementById("loop-notification-panel"), null, "Panel doesn't exist for private browsing");
+
+  win.close();
+});
```
mancas, line 37: @Mardak , could you please review this tests? For some reason I'm getting an error when running them:

Error: Found an unexpected browser window at the end of test run

Standard8, line 37: You might want to switch to:

``` js
let win = yield BrowserTestUtils.openNewBrowserWindow();
yield win.LoopUI.openPanel();
```

BrowserTestUtils has code for waiting for load etc.

You probably also want to change the `win.close()` to a cleanup function at the start of the test - this ensures it gets called even if the test fails, so you're less likely to impact other tests if something goes wrong (and the failures are clearer).

``` js
  registerCleanupFunction(function* () {
    if (win) {
      yield BrowserTestUtils.close(win);
    }
  });
});
```

add-on/chrome/test/mochitest/browser_panel_privateBrowsing.js
```diff
@@ -0,0 +1,26 @@
...
+  yield win.LoopUI.openPanel();
+
+  Assert.equal(win.document.getElementById("loop-notification-panel"), null, "Panel doesn't exist for private browsing");
+});
+
+registerCleanupFunction(function* () {
+  if (win) {
+    yield BrowserTestUtils.closeWindow(win);
+  }
+});
```
mancas, line 37: @Standard8 I don't really know what I'm doing wrong. Both tests seems to work but when I run them, an extra window is being opening. Any ideas?

PR #347, opened 2016-04-08
Bug 1262821 - After exiting a conversation, selecting to "Download Fi…

mancas:
…refox" actually rejoins the conversation


standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -288,12 +288,11 @@ loop.standaloneRoomViews = (function(mozL10n) {
       return (
         <div className="promote-firefox">
           <h2>{mozL10n.get("rooms_promote_firefox_label")}</h2>
-          <button className="btn btn-info"
-                  onClick={this.props.joinRoom}>
+          <a className="btn btn-info" href={loop.config.downloadFirefoxUrl}>
```
Standard8: This should have at least `rel="noreferrer"` added to it (to avoid url/room token leaks), and probably `target="_blank"` as well (like we do with our other links).

PR #348, opened 2016-04-08
Bug 1262455 - Web page title / content information isn't shown in the…

mancas:
… standalone infobar when two people are in the room

### DIFF
--- a/standalone/content/js/standaloneRoomViews.jsx
+++ b/standalone/content/js/standaloneRoomViews.jsx
@@ -527,6 +527,17 @@ loop.standaloneRoomViews = (function(mozL10n) {
       );
     },
 
+    renderWelcomeMessage: function() {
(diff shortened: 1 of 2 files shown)


standalone/content/js/standaloneRoomViews.jsx
```diff
@@ -544,6 +544,15 @@ loop.standaloneRoomViews = (function(mozL10n) {
         case ROOM_STATES.JOINING:
         case ROOM_STATES.JOINED:
         case ROOM_STATES.SESSION_CONNECTED:
+          var roomName = this.props.room.roomName ? this.props.room.roomName :
```
fcampo: just for the sake of code style, I'd take this to a function, same as `renderContext()`

Standard8: Please also include or update tests for the newer functionality.

PR #349, opened 2016-04-08
Bug 1259986 - Make l10n.js handle invalid plural rule values.

Standard8:


daveccrit:
Looks good. r=dcritch

PR #352, opened 2016-04-08
Bug 1262441 - Only display one Welcome message

daveccrit:
r? @Standard8, @dmose

### DIFF
--- a/shared/js/textChatView.jsx
+++ b/shared/js/textChatView.jsx
@@ -202,12 +202,6 @@ loop.shared.views.chat = (function(mozL10n) {
       return (
         <div className={entriesClasses}>
           <div className="text-chat-scroller">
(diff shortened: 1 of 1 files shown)


shared/js/textChatView.jsx
```diff
@@ -199,10 +199,13 @@ loop.shared.views.chat = (function(mozL10n) {
         "text-chat-entries": true
       });
 
+      var hideHelloWelcome = loop.shared.utils.isDesktop() ||
+        this.props.messageList.length > 0;
```
Standard8: This doesn't feel quite right, as the welcome to will go away once a message has been typed or a context tile displayed.

Maybe we should search the message list, or kinda give up and get the room name from the active room store.

PR #356, opened 2016-04-13
Bug 1262540 - Prototype simple en-US string properties completeness test

chrafuse:
@dmose - checks the code for about 72% of the code coverage or finds in code.  Need more work to flesh out a extended search and analysis on missed or unused strings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -492,6 +492,10 @@ eslint:
 flake8: $(VENV)
 	. $(VENV)/bin/activate && flake8 .
 
+.PHONY: check_strings
+check_strings:
+	$(VENV)/bin/python bin/stringsCompletenessTest.py
+
 .PHONY: lint
 lint: eslint flake8
 
--- a/Makefile
+++ b/Makefile
@@ -494,7 +494,7 @@ flake8: $(VENV)
 
 .PHONY: check_strings
 check_strings:
-	$(VENV)/bin/python bin/stringsCompletenessTest.py
+	@$(VENV)/bin/python bin/stringsCompletenessTest.py
 
 .PHONY: lint
 lint: eslint flake8
--- a/bin/stringsCompletenessTest.py
+++ b/bin/stringsCompletenessTest.py
@@ -3,7 +3,6 @@
 ##
 # This script is designed to validate string usage code completeness in the l10n property files and for
 # the standalone loop client and loop add-on executable code.
-# The loop repo is assumed to be https://github.com/mozilla/loop.
 #
 # Run this script from the local version of loop. It assumes that a local
 # version is in directory: ./loop/bin.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -444,6 +444,10 @@ eslint:
 flake8: $(VENV)
 	. $(VENV)/bin/activate && flake8 .
 
+.PHONY: strings
+strings:
```
Standard8: I would suggest changing this to `check_strings`

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+def main(l10n_locale):
+    # capture root path assuming this file is in loop/bin
+    # uncomment the next line if you are not running this from Makefile strings call
+    # os.chdir(os.path.abspath(os.pardir))
+    loop_root_dir = os.getcwd()
+    def_property_dirs = ["add-on", "shared", "standalone"]
+    # loop/locale/en-US
+    def_us_locale_dir = os.path.abspath(os.path.join(loop_root_dir, "locale", l10n_locale))
+    def_property_addon_file_name = os.path.join(def_us_locale_dir, os.extsep.join([def_property_dirs[0], "properties"]))
+    def_property_shared_file_name = os.path.join(def_us_locale_dir, os.extsep.join([def_property_dirs[1], "properties"]))
```
Standard8: You need to run the flake8 / lint tests on this file, currently they fail due to various issues.

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+    def_property_standalone_name = os.path.join(def_us_locale_dir, os.extsep.join([def_property_dirs[2], "properties"]))
+
+    print("l10n_locale", l10n_locale)
+    print(def_us_locale_dir)
+    print(def_property_addon_file_name)
+    print(def_property_shared_file_name)
+    print(def_property_standalone_name)
+
+    # START
```
Standard8: This needs splitting up into at least three functions, something like:
- readStringsFromL10n (read from the l10n files)
- checkStringsInCode
- reportResult

This should make it easier to read and reason about each section

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+    def_property_dirs = ["add-on", "shared", "standalone"]
+    # loop/locale/en-US
+    def_us_locale_dir = os.path.abspath(os.path.join(loop_root_dir, "locale", l10n_locale))
+    def_property_addon_file_name = os.path.join(def_us_locale_dir, os.extsep.join([def_property_dirs[0], "properties"]))
+    def_property_shared_file_name = os.path.join(def_us_locale_dir, os.extsep.join([def_property_dirs[1], "properties"]))
+    def_property_standalone_name = os.path.join(def_us_locale_dir, os.extsep.join([def_property_dirs[2], "properties"]))
+
+    print("l10n_locale", l10n_locale)
+    print(def_us_locale_dir)
+    print(def_property_addon_file_name)
```
Standard8: Most of this debug needs either simplification or removal or not displaying by default. Its too difficult to look at the output at the moment.

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+    print(def_property_standalone_name)
+
+    # START
+
+    l10n_properties = {}
+    l10n_files = [def_property_addon_file_name, def_property_shared_file_name, def_property_standalone_name]
+    # We use encoding="UTF-8" so that we have a known consistent encoding format.
+    # Sometimes the locale isn't always defined correctly for python, so we try
+    # to handle that here.
+    pattern = re.compile(r"^(.*)=")
```
Standard8: I think using `propname_pattern` is probably a bit better here.

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+                for match in re.finditer(pattern, line):
+                    property_key = match.group(1).encode("utf-8")
+                    # print(property_key)
+                    l10n_properties[property_key] = 0
+                    # print(l10n_properties[property_key])
+
+    # directories to look in
+    def_code_dirs = ["add-on/chrome", "add-on/panels/js", "add-on/panels/vendor", "shared/js", "standalone/content/js"]
+    # regex string getter methods to look for
+    pattern = re.compile('_getString\(\"(.*)\"\)|mozL10n\.get\(\"(.*)\"\)')
```
Standard8: `string_match_pattern` ?

Standard8: This also needs breaking up, or explanations for what we're looking for - e.g. see the verbose section example in https://docs.python.org/2/howto/regex.html#compilation-flags

Standard8: Actually, I'd suggest breaking this up into multiple regex. We're likely to want to add more patterns over time, and having the function support an array of regex or something is likely to be beneficial.

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+    # get list of = "literal strings in the code";
+    # Iterate back over code with patterns from not found
+    # or unused strings to find literal strings used and where (line).
+
+    # print report
+    # print (l10n_properties.keys())
+
+    for prop in l10n_properties:
+        print(prop,)
+        print(l10n_properties.get(prop))
```
Standard8: I think the initial report should just print out two things: 1) unused strings, 2) strings not in the locale files.

bin/stringsCompletenessTest.py
```diff
@@ -0,0 +1,140 @@
...
+
+    props_count = len(l10n_properties)
+    print("props_count", props_count)
+    print("props_found", props_found)
+    props_not_used = props_count - props_found
+    props_coverage_perc = (float(props_found)/props_count) * 100
+    print("props_not_used", props_not_used)
+    print("props_coverage_perc %{0:.2f}".format(props_coverage_perc))
+    # END
```
Standard8: We should sys.exit(1) if there's a non-zero amount of unused string/strings not in locale files.

Standard8:
@chris I think this is probably a good starter that we can turn into something we can land. The main thing missing in my comments is working out how we get to 100%, but if we get the script and the output tidied up, then we can figure out how to get there.

PR #360, opened 2016-04-14
Bug 1247894 - add about:loop* URL definitions to add-on

dmose:


add-on/chrome/bootstrap.js
```diff
@@ -1298,6 +1300,10 @@ function startup(data) {
     return;
   }
 
+  // register about: handlers
+  AboutLoop.conversation.register(); // eslint-disable-line no-undef
+  AboutLoop.panel.register(); // eslint-disable-line no-undef
```
Standard8: Rather than doing the `eslint-disable-line`, lets just add AboutLoop to the `add-on/chrome/.eslintrc` like we've done with our other modules.

add-on/chrome/modules/AboutLoop.jsm
```diff
@@ -0,0 +1,93 @@
...
+
+AboutPage.prototype = {
+  QueryInterface: XPCOMUtils.generateQI([Ci.nsIAboutModule]),
+  getURIFlags: function(aURI) { // eslint-disable-line no-unused-vars
+    return Ci.nsIAboutModule.ALLOW_SCRIPT |
+           Ci.nsIAboutModule.HIDE_FROM_ABOUTABOUT |
+           Ci.nsIAboutModule.MAKE_UNLINKABLE |
+           Ci.nsIAboutModule.URI_SAFE_FOR_UNTRUSTED_CONTENT |
+           Ci.nsIAboutModule.URI_MUST_LOAD_IN_CHILD;
+           // XXX note that above line applies to panel and convo! ok?
```
Standard8: This would fix bug 1261350, however, we don't have anything in place yet for handling if the content process crashes - it would need specific handling, otherwise the user would end up with a useless panel.

Hence, I think we should make this only for the conversation window at the moment, and handling fixing the panel in bug 1261350.

add-on/chrome/modules/AboutLoop.jsm
```diff
@@ -0,0 +1,93 @@
...
+  },
+
+  newChannel: function(aURI, aLoadInfo) {
```
Standard8, line 41: I believe something about this function isn't working. I tried a build of FF where I removed the about urls from its core, and got this when trying to open the conversation window:

```
[Exception... "newChannel requires at least one of the 'loadingNode', 'loadingPrincipal', or 'loadUsingSystemPrincipal' properties on the options object."  nsresult: "0x80070057 (NS_ERROR_ILLEGAL_VALUE)"  location: "JS frame :: resource://gre/modules/RemoteAddonsParent.jsm :: AboutProtocolParent.openChannel :: line 268"  data: no]
(unknown)
Empty string passed to getElementById().remote-browser.xml:232:7
TypeError: can't convert undefined to objectRemoteAddonsChild.jsm:258:9
```

If you try just loading `about:panel` in a tab, then you get a similar issue (whereas it loaded before).

I think we need to do something like:

http://hg.mozilla.org/comm-central/file/tip/mail/components/aboutRedirector.js#l56

I haven't quite figured out why the pocket example doesn't have that, nor why `AboutRedirector.cpp` doesn't need to do something similar. So I'm a bit puzzled there.

dmose, line 41: Doing the thing like mail does fixes the newchannel problem, at least.

add-on/chrome/modules/AboutLoop.jsm
dmose: Restructured code a bit so that only about:conversation gets this flag.

dmose:
OK, so, here's an updated version.  It might or might not be right.  Stuff that's going on:

Running against a normal nightly:
- typing about:looppanel in a tab shows nothing, as well as no errors in the console
- typing about:loopconversation in a tab shows this:

[Exception... "newChannel requires at least one of the 'loadingNode', 'loadingPrincipal', or 'loadUsingSystemPrincipal' properties on the options object."  nsresult: "0x80070057 (NS_ERROR_ILLEGAL_VALUE)"  location: "JS frame :: resource://gre/modules/RemoteAddonsParent.jsm :: AboutProtocolParent.openChannel :: line 268"  data: no]
AboutProtocolParent.openChannel()
RemoteAddonsParent.jsm:268
AboutProtocolParent.receiveMessage()
RemoteAddonsParent.jsm:235
TypeError: can't convert undefined to object (RemoteAddonParent.jsm:258:9)

Taking gecko-dev, I then applied your patch to remove the old URLs (though I'm not actually convinced it removes them completely, based on the behavior seen when loading them in tabs with the in-gecko-dev copy of the add-on).

Doing a `make exportc_mc`, rebuilding, and starting up with `mach run` first spews this warning:

1462508352283   addons.xpi  WARN    Problem getting last modified time for /Users/dmose/r/gecko-dev/obj-ff-dbg/dist/NightlyDebug.app/Contents/Resources/browser/features/loop@mozilla.org/chrome/content/shared/js/models.js: [Exception... "Component returned failure code: 0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST) [nsIFile.lastModifiedTime]"  nsresult: "0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST)"  location: "JS frame :: resource://gre/modules/addons/XPIProvider.jsm :: recursiveLastModifiedTime :: line 1911"  data: no] Stack trace: recursiveLastModifiedTime()@resource://gre/modules/addons/XPIProvider.jsm:1911

which suggests the models location move did not get properly adjusted in the export.  I haven't had a chance to poke at this.

Then, about:looppanel behaves reasonably, both in a tab, as well as in the drop-down.

Starting a convo using the drop-down panel, however, doesn't work (the panel doesn't close), but this (possibly unrelated) shows up in the console:

NS_ERROR_MALFORMED_URI: Component returned failure code: 0x804b000a (NS_ERROR_MALFORMED_URI) [nsIIOService2.newURI](PageMetaData.jsm:287)

However, restarting with the about:config interposition pref forced to false does somewhat better.  So there's something shim-my going on here...

Standard8:

PR #367, opened 2016-04-18
Bug 1265430 - Fix gUM Metrics reporting.

Standard8:


shared/js/otSdkDriver.js
```diff
@@ -1001,20 +1001,28 @@ loop.OTSdkDriver = (function() {
...
+        return;
+      }
+
+      this.publisher.off("accessAllowed accessDenied accessDialogOpened streamCreated");
+      this.publisher.destroy();
+      delete this.publisher;
+      delete this._mockPublisherEl;
+
+      this.dispatcher.dispatch(new sharedActions.ConnectionFailure({
+        reason: FAILURE_DETAILS.UNABLE_TO_PUBLISH_MEDIA
```
Mardak, line 1022: Before this PR, this would be dispatched even without a publisher, but now it only dispatches it if there's a publisher to freed up. Is that desired?

shared/js/otSdkDriver.js
```diff
@@ -1001,20 +1001,28 @@ loop.OTSdkDriver = (function() {
...
-        }));
-        this._notifyMetricsEvent("sdk.exception." + error.code + "." + error.message);
+
+      if (error.message && error.message === "DENIED") {
+        // In the DENIED case, this will be handled by _onPublishDenied.
+        return;
       }
+
+      // We free up the publisher here in case the store wants to try
+      // grabbing the media again.
```
Mardak: This comment seems more relevant a few lines lower. Or change the intro as "Only free up the publisher if we have one in case ..."

shared/js/otSdkDriver.js
Mardak, line 1022: Before bug 1210754, this would only be dispatched if event.code is OT.ExceptionCodes.UNABLE_TO_PUBLISH and event.message is "GetUserMedia". Here it seems to happen if there the message isn't DENIED.

Standard8, line 1022: > Before this PR, this would be dispatched even without a publisher, but now it only dispatches it if there's a publisher to freed up. Is that desired?

I think that's fine. When initialising the publisher we get a new object, and we're unlikely to clear it out elsewhere without handling it properly.

Standard8, line 1022: > Before bug 1210754, this would only be dispatched if event.code is OT.ExceptionCodes.UNABLE_TO_PUBLISH and event.message is "GetUserMedia". Here it seems to happen if there the message isn't DENIED.

That's fine, it just means we're actually handling more cases (and providing user feedback) than when we were previously. The denied case is handled in `_onPublishDenied`

PR #368, opened 2016-04-18
Bug 1262441 - String change from welcome to roomname to you have joined roomname

daveccrit:
r? @Standard8, @chrafuse, @mancas

### DIFF
--- a/shared/js/textChatView.jsx
+++ b/shared/js/textChatView.jsx
@@ -53,7 +53,6 @@ loop.shared.views.chat = (function(mozL10n) {
         "received": this.props.type === CHAT_MESSAGE_TYPES.RECEIVED,
         "sent": this.props.type === CHAT_MESSAGE_TYPES.SENT,
         "special": this.props.type === CHAT_MESSAGE_TYPES.SPECIAL,
-        "room-name": this.props.contentType === CHAT_CONTENT_TYPES.ROOM_NAME,
         "text-chat-notif": this.props.contentType === CHAT_CONTENT_TYPES.NOTIFICATION
       });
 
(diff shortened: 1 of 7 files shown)


shared/js/textChatView.jsx
```diff
@@ -202,6 +202,12 @@ loop.shared.views.chat = (function(mozL10n) {
       return (
         <div className={entriesClasses}>
           <div className="text-chat-scroller">
+            {loop.shared.utils.isDesktop() ? null :
+              <p className="welcome-message">
+                {mozL10n.get("rooms_welcome_text_chat_label",
+                  { clientShortname: mozL10n.get("clientShortname2") })}
+              </p>
+            }
```
mancas, line 211: I've tested the patch locally and I can see both welcome messages. I think this lines shouldn't be added again.

shared/js/textChatView.jsx
```diff
@@ -198,11 +198,26 @@ loop.shared.views.chat = (function(mozL10n) {
       var entriesClasses = classNames({
         "text-chat-entries": true
       });
+      var hideHelloWelcome = this.props.messageList.length > 0 ||
+        loop.shared.utils.isDesktop();
+
+      if (this.props.messageList.length > 0) {
+        var msgEntry = this.props.messageList[0];
```
mancas: Is Room name message always the first element of the array? I will suggest to use messageList.filter instead to check if we have a CHAT_CONTENT_TYPES.ROOM_NAME element. @Standard8

Standard8: It probably is generally the first name, but maybe not 100% guaranteed. Having looked at this again, I think we should do a small refactor - I was never too happy with this code (I think I wrote it iirc!), but this is just getting a bit complex and weird now.

I think what we should do is:
- In `textChatStore`, stop storing the room name as part of the message list, instead store it as a separate item in the store.
- As a result, drop the `CHAT_CONTENT_TYPES.ROOM_NAME` type.
- In this file, we reduce down to having the `TextChatRoomName` always rendered at the top of the `text-chat-scroller`
- We pass in `this.state.roomName` (or whatever it gets called) as a prop from `TextChatView` into `TextChatEntriesView` and then into `TextChatRoomName`.
- `TextChatRoomName` then decides if to display the room name or fall back to the shortname.

I think that should simplify the code, especially with removing the special cases, and also help future maintenance.

PR #371, opened 2016-04-20
Bug 1261845 - Wrong background and tooltip for Hello icon during tour

chrafuse:
@Standard8  Added shared loop CSS for slideshow state on the toolbarbutton-badge-stack element.  There was no clear definition of CSS for the background on the icon for any platform so this may fix for all platforms.  Will need to test on all platforms as it works on Ubuntu 14.04 LTS.

Can this be retested by QA again to see if fixed?

### DIFF
--- a/add-on/chrome/bootstrap.js
+++ b/add-on/chrome/bootstrap.js
@@ -539,6 +539,7 @@ var WindowListener = {
           mozL10nId += "-error";
         } else if (this.isSlideshowOpen) {
           state = "slideshow";
+          suffix = ".label";
         } else if (this.MozLoopService.screenShareActive) {
           state = "action";
           mozL10nId += "-screensharing";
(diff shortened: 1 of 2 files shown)


chrafuse:
@Standard8 - Added osx specific background for the #loop-button[state="slideshow"] as it is removed from loop.css seems to be different in the osx button.  I do not have a Mac to inspect the button elements and maybe have slightly different classes or elements.

background: var(--toolbarbutton-hover-background) !important;

Saw this page and thought it may have some insight into platform issues.
https://developer.mozilla.org/en-US/Add-ons/Themes/Common_Firefox_theme_issues_and_solutions

add-on/chrome/skin/shared/loop.css
```diff
@@ -54,11 +54,17 @@
      the same as the hover state to visually indicate that it's in-use.
      */
   #loop-button[state="slideshow"] {
-    background: var(--toolbarbutton-hover-background);
     border-color: var(--toolbarbutton-hover-bordercolor);
     box-shadow: var(--toolbarbutton-hover-boxshadow);
   }
 
+  #loop-button[state="slideshow"] > .toolbarbutton-badge-stack {
```
Standard8: Ok, it appears that overall this still isn't right on Mac. Lets move this part of the css into the platform.css for windows & linux only.

add-on/chrome/skin/osx/platform.css
```diff
@@ -38,4 +38,8 @@
   notification[value="loop-sharing-notification"].paused .notification-button-default {
     background: #fff;
   }
+
+  #loop-button[state="slideshow"] {
```
Standard8: and drop this part of the css.

add-on/chrome/bootstrap.js
```diff
@@ -1422,9 +1423,7 @@ function startup(data) {
     .getService(Components.interfaces.nsIStyleSheetService);
   let sheets = ["chrome://loop-shared/skin/loop.css"];
 
-  if (AppConstants.platform != "linux") {
-    sheets.push("chrome://loop/skin/platform.css");
-  }
+  sheets.push("chrome://loop/skin/platform.css");
```
Standard8: nit: might as well include this in the array on the line above.

PR #373, opened 2016-04-21
No Bug. Add Node 4 to the list of platforms to test on travis.

Standard8:
Since we're now able to support Node 4 to an extent, we should start building against node 4 as well as node 0.10 (until we drop that version). Its unlikely we'll break anything, but turning on both doesn't hurt us much, so its a good additional check.


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

Mardak:
rs=Mardak

PR #329, opened 2016-04-04
Bug 1261488 - prototype window-stapled sidebar using different XUL browsers to display TOC + sidebar or just sidebar

mancas:

PR #376, opened 2016-04-22
Bug 1261488 - ToC & sidebar proto-impl (rebased & unforked)

dmose:
This is the first part of doing the unforking and untangling that @Standard8 requested in his review of https://github.com/mozilla/loop/pull/329.


dmose:
Interestingly, when I run `make test` locally, I get different unit test failures (in standlone) than the ones in either of the travis builds (in conversation.jsx).  Even more interestingly, running the unit tests on 45.0 Mac has them all passing.  At least the Travis stuff probably needs to be fixed before landing.

dmose:
Also, the UI-showcase is in relatively bad shape (a lot of views are broken).  I suspect this doesn't want to block landing, but probably does want to be one of the first things we fix post-landing.

add-on/chrome/bootstrap.js
```diff
@@ -1431,6 +1473,11 @@ function shutdown(data, reason) {
   // Stop waiting for browser windows to open.
   wm.removeListener(WindowListener);
 
+  // unregister about: handlers
+  AboutLoop.conversation.unregister(); // eslint-disable-line no-undef
+  AboutLoop.panel.unregister(); // eslint-disable-line no-undef
+  AboutLoop.toc.unregister(); // eslint-disable-line no-undef
```
dmose, line 1479: Hmm, I think this was mentioned in Mark's original review comments.  Do we still need to go through and address his original comments now that we've unforked?

shared/js/activeRoomStore.js
```diff
@@ -293,6 +293,7 @@ loop.store.ActiveRoomStore = (function(mozL10n) {
       var roomToken = this._storeState.roomToken;
       loop.request("Rooms:PushSubscription", ["delete:" + roomToken, "update:" + roomToken]);
       loop.subscribe("Rooms:Delete:" + roomToken, this._handleRoomDelete.bind(this));
+      loop.subscribe("Rooms:Delete:" + roomToken, this._handleRoomDelete.bind(this));
```
dmose: This addition looks like an inadvertent line duplication?

shared/js/activeRoomStore.js
```diff
@@ -362,6 +363,7 @@ loop.store.ActiveRoomStore = (function(mozL10n) {
      *                   if Firefox handles the room or not.
      */
     fetchServerData: function(actionData) {
+      console.info(actionData);
```
dmose: This wants to go away, I bet.  :-)

shared/js/toc.jsx
```diff
@@ -0,0 +1,559 @@
...
+                {this.props.tile.description}
+            </a>
+            <h3 className="tile-url">{this.props.tile.location}</h3>
+          </div>
+        </div>
+      );
+    }
+  });
+
+  var SidebarView = React.createClass({
```
dmose, line 313: If I recall correctly, wasn't the plan after Mark's review to split this off into its own file, since it's not really part of the ToC itself?

add-on/chrome/bootstrap.js
mancas, line 1479: I can't see the comment you're talking about

https://github.com/mozilla/loop/pull/329

mancas:
I did squash of the latests commits to get the unit tests stuff all together in one commit

add-on/chrome/bootstrap.js
Standard8, line 1479: I think @dmose is thinking about the comments on #360. I'm happy for those comments to be addressed separately there, landed on master, and then de-forked when merging to akita.

add-on/chrome/bootstrap.js
```diff
@@ -347,6 +349,41 @@ var WindowListener = {
...
+        let ownerDocument = gBrowser.ownerDocument;
+        var browser = ownerDocument.getElementById("browser");
+
+        let sidebarBrowser = document.createElementNS(kNSXUL, "browser");
+        sidebarBrowser.setAttribute("id", "loop-side-browser");
+        sidebarBrowser.setAttribute("disable-history", "true");
+        sidebarBrowser.setAttribute("disable-global-history", "true");
+
+        // XXX something like these seem likely to be required once we
+        // electrolyze this along with a URI_MUST_LOAD_IN_CHILD in
```
Standard8, line 367: The sidebar is already URI_MUST_LOAD_IN_CHILD. I think this probably doesn't apply, however, lets change the `XXX` to `XXX akita` and verify it before too much longer.

Actually I think having the must load in child forces it into a remote browser, but specifying the browser element remote by default optimises the process.

So just update the `XXX` for now please.

add-on/chrome/modules/AboutLoop.jsm
```diff
@@ -0,0 +1,114 @@
...
+
+XPCOMUtils.defineLazyGetter(AboutLoop, "toc", () => {
+  return new AboutPage("chrome://loop/content/panels/toc.html",
+                       "looptoc",
+                       "A1220CE0-E5D1-45B6-BEBA-3706166A2AA4",
+                       "About Loop ToC",
+                       Ci.nsIAboutModule.ALLOW_SCRIPT |
+                        Ci.nsIAboutModule.HIDE_FROM_ABOUTABOUT |
+                        Ci.nsIAboutModule.URI_SAFE_FOR_UNTRUSTED_CONTENT |
+                        Ci.nsIAboutModule.MAKE_UNLINKABLE); // XXX load in child?
```
Standard8: `XXX` -> `XXX akita-sidebar`

add-on/panels/js/sidebar.jsx
```diff
@@ -0,0 +1,170 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Standard8, line 1: I think this still needs to be a `git mv` from conversation.jsx.

add-on/panels/sidebar.html
```diff
@@ -49,8 +50,7 @@
     <script type="text/javascript" src="shared/js/textChatView.js"></script>
     <script type="text/javascript" src="shared/js/linkifiedTextView.js"></script>
     <script type="text/javascript" src="panels/js/desktopViews.js"></script>
-    <script type="text/javascript" src="panels/js/feedbackViews.js"></script>
```
Standard8, line 52: Please add an `XXX Akita` for the feedback views. This is something I've just realised the UX designs are probably missing.

shared/css/sidebar.css
```diff
@@ -0,0 +1,98 @@
+/* http://meyerweb.com/eric/tools/css/reset/
+   v2.0 | 20110126
+   License: none (public domain)
```
Standard8: This is the wrong license, we should be using one of the MPL headers here: https://www.mozilla.org/en-US/MPL/headers/

shared/css/toc.css
```diff
@@ -0,0 +1,215 @@
+.standalone-room-wrapper .toc-wrapper {
```
Standard8, line 6: This file should get an MPL license header as well.

shared/js/toc.jsx
Standard8, line 313: xref: https://github.com/mozilla/loop/pull/329/files#r60665491

I think it would be good to de-fork this now, or to simplify things, it could be the first thing we land after this PR.

dmose, line 313: Just talked to Mark, this will be fixed by doing the git mv he suggests.

add-on/chrome/bootstrap.js
dmose, line 1479: Yeah, I got confused; never mind!

shared/js/toc.jsx
dmose, line 313: I now think Mark and I got confused when we talked.  There are two SidebarViews, one that is shared and embedded standalone view, and one which is the desktop view that embeds the shared view.  These comments refer to the shared view.  

It may be that we want to de-fork one or both views now; I'm trying to trace the lineage of them.

dmose, line 313: OK, so, loop.shared.toc.SidebarView is the embedded version, and is being commented on in this part of the PR.  It is a fork of standaloneRoomView.jsx:StandaloneRoomView.

dmose, line 313: Effectively, there are now two forks of StandaloneRoomView: StandaloneRoomView itself, and loop.shared.toc.SidebarView.  The functionality of the old StandaloneRoomView is to some degree being split into the two objects.  The split happened by forking, not by cleanly decoupling the relevant pieces. 

Ways forward that I can imagine:

a) factor out the common subset of code and make it a mixin, or, even better, a composable 
higher-order component.  Either of these sounds like a lot of work, without a tremendous amount of gain.

b) very soon, go through each of the components and see what we can rip out

c) over time, notice the pieces which aren't needed in each component and remove them as we notice them

I don't think any of these should block landing to akita.

I'm not sure what we want to do here; either b) or c) sounds reasonable to me, and I'm inclined towards c.  @Standard8, what do you think?

dmose, line 313: That said, I do think loop.shared.toc.SidebarView should move to its own file and not be exposed at the loop.shared.toc part of the loop API hierarchy.  I think that can be an XXX akita-sidebar comment, however.

dmose, line 313: Note, also, that git does not have a conceptual thing like `hg cp`, which is I'm not suggesting something like that.  See http://stackoverflow.com/questions/36848705/git-equivalent-to-hg-cp?answertab=votes#tab-top for details.

dmose, line 313: I'm leaving this alone for now.

add-on/panels/js/sidebar.jsx
dmose, line 1: OK, rebased with that all set up.

add-on/panels/sidebar.html
dmose, line 52: Added.

shared/css/toc.css
dmose, line 6: The file actually has it now.  I'm not sure when that got added, but we're all good here.

PR #380, opened 2016-04-25
Bug 1225057 - Rewrite NotificationCollection/NotificationModel to be a store

malayaleecoder:
Part 1 of the bug. 
@Standard8 : Please have a look.

### DIFF
--- a/add-on/panels/panel.html
+++ b/add-on/panels/panel.html
@@ -23,7 +23,7 @@
 
     <script type="text/javascript" src="shared/js/loopapi-client.js"></script>
     <script type="text/javascript" src="shared/js/utils.js"></script>
-    <script type="text/javascript" src="shared/js/models.js"></script>
+    <script type="text/javascript" src="panels/js/models.js"></script>
     <script type="text/javascript" src="shared/js/mixins.js"></script>
     <script type="text/javascript" src="shared/js/actions.js"></script>
     <script type="text/javascript" src="shared/js/dispatcher.js"></script>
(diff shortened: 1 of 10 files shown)


Standard8:
It looks like you missed a change in `panel.html` for updating the directory of `models.js` - when I did a `make test` locally the functional tests failed (at the moment the tests aren't run on Travis that reports to the PR).

Can you also change the commit message to be something like `Bug 1225057 - Part 1. Move the notification models and views into the desktop specific directory. r=Standard8`

I think once that is done, we should be there!

malayaleecoder:
@Standard8 : I noticed the Error on `make test` but thought that it would be due to the fact that it was not able to open Chrome. Anyway, I have made the changes, but still observe the same error. Please have a look.

add-on/panels/panel.html
```diff
@@ -23,7 +23,7 @@
 
     <script type="text/javascript" src="shared/js/loopapi-client.js"></script>
     <script type="text/javascript" src="shared/js/utils.js"></script>
-    <script type="text/javascript" src="shared/js/models.js"></script>
+    <script type="text/javascript" src="add-on/panels/js/models.js"></script>
```
Standard8: This needs to be just `panels/js/models.js` (see the paths for some other panels files further down in the file).

malayaleecoder:
@Standard8 Sorry, did not notice that in a hurry.

Standard8:
That's great. r=Standard8

PR #387, opened 2016-04-27
Bug 1267960 - Hello popup notification is displayed upon coping the URL bar even if the user is already in a Hello call

Mardak:


add-on/chrome/bootstrap.js
```diff
@@ -469,6 +469,11 @@ var WindowListener = {
           // Do the normal behavior first.
           controller._doCommand.apply(controller, arguments);
 
+          // Don't bother prompting the user if already sharing.
+          if (this.MozLoopService.screenShareActive) {
```
fcampo, line 481: to fullfill all the conditions, I think we should check also that:
- `loop.copy.ticket != -1`, that shows that Ctrl-C has been used before
- `roomList.length != 0`, that shows that a call has been done before (no matter if currently ongoing)

Mardak, line 481: No need to check `ticket` as `addCopyPanel` only gets called if ticket passes throttling checks, so `ticket` will be 255. I did ask about not showing the panel if the user has ever successfully created a room, and it wasn't written in the bug, but it was explicitly mentioned to only check for an active session to prevent copy panel and not previous rooms.

daveccrit:
Looks good, r=dcritch

PR #390, opened 2016-04-28
Bug 1268397 - Don't open the conversation window after creating a new…

mancas:
… room

### DIFF
--- a/add-on/chrome/modules/LoopRooms.jsm
+++ b/add-on/chrome/modules/LoopRooms.jsm
@@ -55,6 +55,9 @@ const roomsPushNotification = function(version, channelID) {
   return LoopRoomsInternal.onNotification(version, channelID);
 };
 
+// Table of Contents url
+const TOC_URL = "about:looptoc#";
+
 // Since the LoopRoomsInternal.rooms map as defined below is a local cache of
 // room objects that are retrieved from the server, this is list may become out
 // of date. The Push server may notify us of this event, which will set the global
@@ -730,8 +733,9 @@ var LoopRoomsInternal = {
     }
   },
 
+  // XXX akita Bug 1268826 tidy up chat window code
   open: function(roomToken) {
-    let windowData = {
+    /* let windowData = {
       roomToken: roomToken,
       type: "room"
     };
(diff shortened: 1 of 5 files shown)


add-on/panels/js/roomStore.js
```diff
@@ -544,7 +550,8 @@ loop.store = loop.store || {};
      */
     openRoom: function(actionData) {
       loop.requestMulti(
-        ["Rooms:Open", actionData.roomToken],
+        ["OpenURL", TOC_URL + actionData.roomToken],
+        ["LoadSidebar", actionData.roomToken],
```
Standard8: I think these should really be in `LoopRooms#open`. The reason is, that as we move to a new window, we're going to want to make sure that we only have one instance of a room open at any one time. The only place we can really do that is from the privileged code.

So I think it makes sense to keep using that function for now (I'm not expecting the one-instance code to be written yet).

I realise there's some XXX code for `LoopRooms#open`, maybe just comment out the old code within the function and add the XXX akita to tidy up all chat window associated code later (maybe file a bug at the same time?)

mancas: @Standard8 then what if we move the LoadSidebar code into the LoopRooms#open? Since we cannot load a different room in the same window, there's no reason to have an exposed method. wdyt?

Standard8: I think that would probably be fine. There's a chance with moving to a dedicated room that we'll also need to have some sort of sidebar open method, but until we've worked out what we're doing its hard to say.

So yes, doing the LoadSidebar from LoopRooms#open, seems reasonable, we can always move it again later if necessary.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -730,8 +733,9 @@ var LoopRoomsInternal = {
     }
   },
 
+  // XXX akita Bug 1268826 tidy up chat window code
   open: function(roomToken) {
-    let windowData = {
+    /* let windowData = {
```
Standard8, line 738: Just a thought: This change is going to break some of the xpcshell-tests. I'm ok with that, but maybe add an XXX to test_looprooms.js to fix after the separate window has landed?

PR #407, opened 2016-05-04
Bug 1268811 - Refactor code and remove akita-sidebar comments in side…

mancas:
…bar.jsx


add-on/panels/js/sidebar.jsx
```diff
@@ -119,17 +119,34 @@ loop.sidebar = (function(mozL10n) {
...
+          callback(null, currGuid);
+        },
+        set: function(guid, callback) {
+          // See nsIPrefBranch
+          var PREF_STRING = 32;
+          currGuid = guid;
+          loop.request("SetLoopPref", "ot.guid", guid, PREF_STRING);
+          callback(null);
+        }
+      });
```
mancas, line 132: Not sure if we need to keep this. @Standard8

add-on/panels/js/sidebar.jsx
```diff
@@ -148,22 +165,21 @@ loop.sidebar = (function(mozL10n) {
...
                         dispatcher={dispatcher} />, document.querySelector("#main"));
 
-      var locationData = sharedUtils.locationData();
-      var hash = locationData.hash.match(/#(.*)/);
-
       dispatcher.dispatch(new sharedActions.SetupWindowData({
-        windowId: "id-test",
-        roomToken: hash[1],
+        roomToken: roomToken,
         type: "room"
```
mancas: @Standard8 the "type" field is really needed or it's part of the old video calls?

add-on/panels/js/sidebar.jsx
Standard8, line 132: Yes we do need to keep this (desktop only). You could probably change the comment to end with something like "don't work in about: pages".

add-on/panels/js/sidebar.jsx
Standard8: I think we can drop it. It was when we had the call vs room mechanisms, and we're unlikely to need that soon.

add-on/panels/test/sidebar_test.js
```diff
@@ -191,165 +171,32 @@ describe("loop.conversation", function() {
...
-        "remoteCursorClick": false
-      });
-
-      sinon.assert.notCalled(clickRemoteCursorStub);
-    });
-
-    // XXX akita-sidebar
-    it.skip("should display the RoomView for rooms", function() {
-      conversationAppStore.setStoreState({ windowType: "room" });
+    it("should display the DesktopSidebarView", function() {
```
Standard8, line 197: We should also add a test for "ROOM_STATE.GATHER" since I believe that can happen, even if its only for a very short time.

Now I think about it, its probably best to also change the GATHER state just to display nothing, rather than the text, to avoid any possible flicker on slower machines.

add-on/panels/js/sidebar.jsx
```diff
@@ -148,22 +165,20 @@ loop.sidebar = (function(mozL10n) {
...
+        roomToken: roomToken
       }));
+
+      loop.request("TelemetryAddValue", "LOOP_ACTIVITY_COUNTER", constants.LOOP_MAU_TYPE.OPEN_CONVERSATION);
     });
   }
 
   return {
-    init: init
+    DesktopSidebarView: DesktopSidebarView,
```
Standard8, line 175: Please can you either add an XXXakita  and empty `describe("DesktopSidebarView", function() { it.skip("should have tests")})` in the tests, or write tests for this view please.

PR #411, opened 2016-05-04
Bug 1257266 - Change NPS survey link to dynamically add firefox hello version number to survey link

daveccrit:
r? @Standard8, @dmose


add-on/chrome/modules/MozLoopService.jsm
```diff
@@ -1587,6 +1587,18 @@ this.MozLoopService = {
   },
 
   /**
+   * Returns the addon version
+   *
+   * @return {String} A string containing the Addon Version
+   */
+  get addonVersion() {
```
Standard8, line 1594: We should add an xpcshell test for this function. However, since we also want to get this out fairly quick, I'm happy for that to be done in a follow-up.

PR #412, opened 2016-05-05
Bug 1270020 - Refactor code and remove akita-sidebar comments in Medi…

mancas:
…aLayoutView and MediaView


shared/js/textChatView.jsx
```diff
@@ -442,9 +441,6 @@ loop.shared.views.chat = (function(mozL10n) {
             dispatcher={this.props.dispatcher}
             showPlaceholder={!hasSentMessages}
             textChatEnabled={this.state.textChatEnabled} />
-          <sharedViews.AdsTileView
```
Standard8, line 445: The component should probably get an XXX statement for akita saying its currently unused & we either need to remove it or use it (maybe get a bug on file as well, as I don't think the requirement for the ToC is tracked too well).

shared/js/views.jsx
```diff
@@ -977,67 +935,39 @@ loop.shared.views = (function(_, mozL10n) {
...
-                mediaType="remote"
-                posterUrl={this.props.remotePosterUrl}
-                srcMediaElement={this.props.remoteSrcMediaElement} />
-              {this.state.localMediaAboslutelyPositioned ?
-                this.renderLocalVideo() : null}
-              {this.props.children}
-            </div>
+            {
+              this._shouldShowRemoteStream() ?
+              <div className={remoteStreamClasses}>
```
Standard8: This feels like its getting a bit confusing and difficult to maintain. I can see there being some hidden gotchas when it comes to css as well.

The main reason previously for moving the elements around was due to the different screen layouts and requirements (e.g. not always showing the screen share). As we've now dropped some of that complication, I'm thinking we should be able to simplify this. Maybe something like:

``` js
<MediaView mediaType="remote" ... />
<MediaView mediaType="local" ... />
<MediaButtonsView .../>
<MediaWaitView ... />
```

I'm not sure, but I'm also wondering about whether or not we should move the decision of whether to render the contents of a component into the component itself, I think it would be nice, but I'm not sure if that would make sense or not. Maybe try it out, or see if @dmose has some thoughts?

shared/css/sidebar.css
```diff
@@ -3,96 +3,502 @@
...
+  width: 250px;
+  height: 100%;
+}
+
+.sidebar-wrapper > .media-layout {
+  /* 188px video height + 40px media buttons height */
+  height: calc(188px + 40px);
+}
+
+.sidebar-wrapper > .media-layout > .media-wrapper {
```
Standard8: This feels a bit strange, reaching down into the wrapper, when you're not doing that elsewhere.

Standard8:
General note: we should start thinking about updating the ui-showcase as well.

shared/js/textChatView.jsx
mancas, line 445: @Standard8 FYI, The requirement is now tracked in Aha!:
https://mozilla.aha.io/features/HAO-85#comment_6282609811081516190

shared/css/conversation.css
```diff
@@ -107,38 +107,11 @@ html[dir="rtl"] .conversation-toolbar-media-btn-group-box > button:first-child:h
   content: url("../img/audio.svg");
 }
 
-.btn-mute-audio.muted:after {
-  content: url("../img/audio-mute.svg");
```
Standard8, line 111: I believe all of the svgs listed below can be removed as part of removing the css:
- `audio-mute.svg`
- `audio-hover.svg`
- `audio-mute-hover.svg`
- `video-hover.svg`
- `video-mute.svg`
- `video-mute-hover.svg`

shared/css/sidebar.css
```diff
@@ -3,96 +3,572 @@
...
+  background-image: url("../img/audio-call-avatar.svg");
+  background-repeat: no-repeat;
+  background-color: #4BA6E7;
+  background-size: contain;
+  background-position: center;
+  /*
+   * Expand to fill the available space, since there is no video any
+   * intrinsic width. XXX should really change to an <img> for clarity
+   */
+  height: calc(100% - 40px);
```
Standard8: What's the 40px for? It seems to adjust the height of the avatar when muted, so you get this strange resizing & border.

shared/css/sidebar.css
```diff
@@ -3,96 +3,572 @@
...
+/*------------------------------------*\
+    TEXT CHAT
+\*------------------------------------*/
+
+.text-chat-view {
+  flex: 0 0 auto;
+  /* Leaves enough room for the streams and the media buttons */
+  height: calc(100% - 228px);
+  /* Text chat is a fixed 272px width for normal displays. */
+  width: 250px;
```
Standard8, line 194: The value and comment are different sizes... maybe just drop the size in the comment?

shared/js/views.jsx
```diff
@@ -911,73 +920,63 @@ loop.shared.views = (function(_, mozL10n) {
...
             mediaType="local"
             posterUrl={this.props.localPosterUrl}
             srcMediaElement={this.props.localSrcMediaElement} />
+          <MediaButtonsView
+            audio={this.props.audio}
+            dispatcher={this.props.dispatcher}
+            leaveRoom={this.props.leaveRoom}
+            video={this.props.video} />
+          <MediaWaitView
+            showMediaWait={this.props.showMediaWait} />
```
Standard8, line 964: I'm not seeing the media wait view on the standalone, when I'm being prompted for gUM. Did something break?

mancas, line 964: @Standard8 I'm not able to see now the loading view when you accept the gUM prompt but the component is updated so quickly and maybe that's the reason

Standard8, line 964: Yeah, it will do for the local view.

PR #414, opened 2016-05-05
Bug 1270525 - Only for feedback, don't merge; Use babel to process all js files except bootstrap

fzzzy:
I will also add jsm files and bootstrap.js to this, now that @Standard8 has said we should try it


mozilla-autolander-deprecated:

Mardak:
In #415, Makefile ends up with just one recipe that builds with babel. That single recipe could be reused for this PR to clean things up.

Additionally, if we can be somewhat loose with our prerequisites by making sure we don't have both a .js and .jsx file in the same directory, we could do something like `$(BUILT)/ui/%.js: PREREQ = ui/$(@F)*` (note the \* after `$(@F)` where that will set PREREQ to any .js\* for the matching file, e.g., for built/ui/fake-l10n.js => PREREQ = ui/fake-l10n.js and for built/ui/ui-showcase.js => PREREQ = ui/ui-showcase.jsx.

We could even use that same single recipe to build .jsm files if we're okay with the output being .js (where we would need to update import references to use `modules/MozLoopAPI.js` instead.)

fzzzy:
#415 looks really good to me, although I guess it's really up to @Standard8 whether this makes the makefile too hard to understand for make newbies. In my opinion, even though it uses some more advanced make constructs, it removes so much code that I think it is worth it.

PR #415, opened 2016-05-06
Clean up Makefile to use a single shared babel recipe for building js files.

Mardak:
This would probably simplify #414 quite a bit. r?@dmose

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,34 +233,18 @@ $(BUILT)/add-on/install.rdf: add-on/install.rdf.in \
 	         -o $@ $<
 
 
-# Build our jsx files into appropriately placed js files.  Note that the rules
-# are somewhat repetitive, and less elegant than they might be.  We _could_
-# use more terse make syntax here, but the idea is that it's better to
-# keep the feel readable/maintainable by non Make-experts, even at the cost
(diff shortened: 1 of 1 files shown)


mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

mozilla-autolander-deprecated:

Mardak:
There was some discussion to get @Standard8's feedback given the original comment: `the idea is that it's better to keep the feel readable/maintainable by non Make-experts, even at the cost of some repetition.`

Standard8:
> There was some discussion to get @Standard8's feedback given the original comment: `the idea is that it's better to keep the feel readable/maintainable by non Make-experts, even at the cost of some repetition.`

I'm fine with changing. I think its better to let this be tidied up a bit, rather than struggle on with limitations. The documentation in the patch helps understanding as well, even if people aren't too familiar with make.

Standard8:
r=Standard8 with a comment added about the ordering of the PREREQ lines.

mozilla-autolander-deprecated:

PR #416, opened 2016-05-06
Bug 1270778 - Trim Base64 padding from FxA relier keys

ghost:
A workaround until the FxA content server removes padding from keys. Sorry for the trouble!


ghost:
Filed mozilla/fxa-content-server#3713 to make sure the server returns unpadded strings for `kBr`, too.

Standard8:
Thank you for the fixes!

PR #418, opened 2016-05-09
Bug 1247894 - about:loop* update

dmose:
This PR attempts to back out the old version of the about:loop branch from the akita patch, and land the most up-to-date version.  There's at least one problem where the sidebar doesn't load that needs to be tracked down, plus some manual testing to make sure nothing else got broken is necessary.

### DIFF
--- a/add-on/chrome/bootstrap.js
+++ b/add-on/chrome/bootstrap.js
@@ -16,6 +16,8 @@ const CURSOR_CLICK_DELAY = 1000;
 // Due to bug 1051238 frame scripts are cached forever, so we can't update them
 // as a restartless add-on. The Math.random() is the work around for this.
 const FRAME_SCRIPT = "chrome://loop/content/modules/tabFrame.js?" + Math.random();
+// Process scripts may or may not need this too, but it won't hurt us
+const PROCESS_SCRIPT = "chrome://loop/content/modules/loop-content-process.js?" + Math.random();
 
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
@@ -323,8 +325,8 @@ var WindowListener = {
         // expected errors to the console, so we catch them here.
         this.MozLoopService.initialize(WindowListener.addonVersion).catch(ex => {
           if (!ex.message ||
-              (!ex.message.contains("not enabled") &&
-               !ex.message.contains("not needed"))) {
+              (!ex.message.includes("not enabled") &&
+               !ex.message.includes("not needed"))) {
             console.error(ex);
           }
         });
@@ -1432,10 +1434,9 @@ function startup(data) {
     return;
   }
 
-  // register about: handlers
-  AboutLoop.conversation.register(); // eslint-disable-line no-undef
-  AboutLoop.panel.register(); // eslint-disable-line no-undef
-  AboutLoop.toc.register(); // eslint-disable-line no-undef
+  // Take care of things that need to happen in each process, such as
+  // registering about: handlers
+  Services.ppmm.loadProcessScript(PROCESS_SCRIPT, true);
 
   createLoopButton();
 
@@ -1508,10 +1509,17 @@ function shutdown(data, reason) {
   // Stop waiting for browser windows to open.
   wm.removeListener(WindowListener);
 
-  // unregister about: handlers
-  AboutLoop.conversation.unregister(); // eslint-disable-line no-undef
-  AboutLoop.panel.unregister(); // eslint-disable-line no-undef
-  AboutLoop.toc.unregister(); // eslint-disable-line no-undef
+  let ppmm = Cc["@mozilla.org/parentprocessmessagemanager;1"]
+           .getService(Ci.nsIMessageBroadcaster);
+  ppmm.broadcastAsyncMessage("LoopShuttingDown");
+
+  // Don't bother listening for a success response,
+  // there's nothing we could do if it failed...
+  AboutLoop.conversation.unregister();
+  AboutLoop.panel.unregister();
+  AboutLoop.toc.unregister();
+
+  Services.ppmm.removeDelayedProcessScript(PROCESS_SCRIPT);
 
   // If the app is shutting down, don't worry about cleaning up, just let
   // it fade away...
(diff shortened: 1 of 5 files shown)


add-on/chrome/bootstrap.js
```diff
@@ -1404,10 +1404,8 @@ function startup(data) {
     return;
   }
 
-  // register about: handlers
-  AboutLoop.conversation.register(); // eslint-disable-line no-undef
-  AboutLoop.panel.register(); // eslint-disable-line no-undef
-  AboutLoop.toc.register(); // eslint-disable-line no-undef
+  Services.ppmm.loadProcessScript(
+    "chrome://loop/content/modules/loop-content-process.js", true);
```
Standard8: According to [devmo](https://developer.mozilla.org/en-US/Firefox/Multiprocess_Firefox/Message_Manager/Process_scripts) process scripts work the same way as frame scripts. Can you therefore set this up with a const url in a similar way to FRAME_SCRIPT please?

That will help us when we get to proper restartless.

add-on/chrome/bootstrap.js
```diff
@@ -1480,10 +1478,22 @@ function shutdown(data, reason) {
   // Stop waiting for browser windows to open.
   wm.removeListener(WindowListener);
 
-  // unregister about: handlers
-  AboutLoop.conversation.unregister(); // eslint-disable-line no-undef
-  AboutLoop.panel.unregister(); // eslint-disable-line no-undef
-  AboutLoop.toc.unregister(); // eslint-disable-line no-undef
+  // XXX do we really want the global PMM here, since it (unnecessarily, I
+  // think) sends a message to the CIPMM (is there any alternative?)
```
Standard8: The way I understand it, the global PMM is the right thing to do. We could potentially scope it down to PPMMs and see if we can find out which ones are attached to which browsers (so for example about:loop\* would only work in the panel, sidebar & toc areas), however I think we don't really need that limitation.

add-on/chrome/bootstrap.js
```diff
@@ -1480,10 +1478,22 @@ function shutdown(data, reason) {
...
+  let ppmm = Cc["@mozilla.org/parentprocessmessagemanager;1"]
+           .getService(Ci.nsIMessageBroadcaster);
+  ppmm.broadcastAsyncMessage("LoopShuttingDown");
+
+  // XXX should we be listening for a response?  I'm not sure what we'd do if
+  // we got one, so it doesn't really seem worth it
```
Standard8: I don't think its worth it. The scripts won't unload as such, they'll just unregister appropriately.

add-on/chrome/modules/loop-content-process.js
```diff
@@ -0,0 +1,61 @@
...
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("chrome://loop/content/modules/AboutLoop.jsm");
+
+function AboutLoopChildListener() {
+}
+AboutLoopChildListener.prototype = {
+  onStartup: function onStartup() {
+
+    if (Services.appinfo.processType ==
+        Services.appinfo.PROCESS_TYPE_CONTENT) {
```
Standard8, line 23: Is the reason this doesn't apply to the chrome process, the fact the messages don't get sent to the parent process? If so, it'd be nice to add a comment here to that affect.

add-on/chrome/bootstrap.js
```diff
@@ -1480,10 +1478,22 @@ function shutdown(data, reason) {
...
+  AboutLoop.conversation.unregister();
+  AboutLoop.panel.unregister();
+  AboutLoop.toc.unregister();
+
+  // XXX this probably races with the broadcastAsyncMessage above, but it'd
+  // be a bit surprising if that actually matters
+  Services.ppmm.removeDelayedProcessScript(
```
Standard8: There's no race here - process scripts don't get removed on shutdown (same as for frame scripts).

add-on/chrome/modules/AboutLoop.jsm
```diff
@@ -77,38 +78,39 @@ var AboutLoop = {};
...
+                "A5DE152B-DE58-42BC-A68C-33E00B17EC2C",
+                "About Loop Panel",
+                Ci.nsIAboutModule.ALLOW_SCRIPT |
+                Ci.nsIAboutModule.HIDE_FROM_ABOUTABOUT |
+                Ci.nsIAboutModule.MAKE_UNLINKABLE |
+                Ci.nsIAboutModule.URI_SAFE_FOR_UNTRUSTED_CONTENT)
+);
 
 XPCOMUtils.defineLazyGetter(AboutLoop, "toc", () => {
   return new AboutPage("chrome://loop/content/panels/toc.html",
```
Standard8: I think you might as well change the style of this one to match the other two that you've changed.

add-on/chrome/modules/loop-content-process.js
```diff
@@ -0,0 +1,61 @@
...
+      // XXX can or should use weak?
+      Services.cpmm.addMessageListener("LoopShuttingDown", this, true);
```
Standard8, line 25: It isn't going to matter, since this script never goes away.

add-on/chrome/bootstrap.js
dmose: Done.  It's not obvious to me whether the randomization hack is needed, but I put it in anyway since it can't really hurt us.

add-on/chrome/bootstrap.js
dmose: OK, comment removed.

add-on/chrome/bootstrap.js
dmose: Is this true even in a restartless world, where someone enables, disables, and enables again?

add-on/chrome/modules/loop-content-process.js
dmose, line 23: Done.

add-on/chrome/modules/AboutLoop.jsm
dmose: Done.

add-on/chrome/modules/loop-content-process.js
dmose, line 25: Comment removed.

add-on/chrome/bootstrap.js
dmose: OK, comment removed.

dmose:
Bah, I think I incorporated and/or pushed the review comment fixes into the wrong branch.  Will clean up tomorrow.

add-on/chrome/bootstrap.js
Standard8: Well, its an async message, and what can you do about it if it doesn't get there and back? Unless we block on the response, but then we might hang the main process... I think this is just one of the instances where we have to trust it gets there.

PR #419, opened 2016-05-09
Bug 1271210 - Add screen share button to the sidebar which allow the …

mancas:
…user to enable/disable it


shared/img/screen.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="15px" height="14px" viewBox="0 0 15 14" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 3.7.2 (28276) - http://www.bohemiancoding.com/sketch -->
+    <title>Screen Share Active</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Room-Description---Hello-Window" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Room---Screen-Sharing-Enabled" transform="translate(-1319.000000, -143.000000)" fill="#4C4C4C">
```
Mardak: Where did these images come from? Do we want the fill color to match up with the blue that's used in the other buttons? Also, we should `svgo` these svg files.

shared/css/conversation.css
```diff
@@ -139,6 +139,19 @@ html[dir="rtl"] .conversation-toolbar-media-btn-group-box > button:first-child:h
...
+}
+
+
+
+.btn-mute-screen.muted:after {
+  content: url("../img/screen-mute.svg");
+}
+
+
+
```
Mardak: Lots of extra newlines here. Also are we expecting a hover state?

shared/img/screen.svg
mancas: @Mardak these images come from Pau. I don't think the color of the icons must be a blocker since we haven't the visuals yet, and probably we should change colors and the visual aspect of lot of components/images.

I'm going to svgo the files.

mancas: Perhaps it's a good idea to add a XXX akita comment to apply better style to the button

shared/js/views.jsx
```diff
@@ -301,6 +301,40 @@ loop.shared.views = (function(_, mozL10n) {
...
+      };
+    },
+
+    toggleScreenShare: function() {
+      if (this.props.disabled) {
+        return;
+      }
+
+      var action = this.props.muted ? "StartBrowserShare" : "EndScreenShare";
+
```
Mardak: nit: drop the newline between the `var` and `dispatch`.

PR #422, opened 2016-05-10
Bug 1270525 - Use babel on all js files

fzzzy:
Switch to using simplified makefile with make's second pass

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
-  "plugins": [
-    "transform-react-display-name",
-    "transform-react-jsx"
+  "presets": [
+    "react",
+    "es2015"
   ]
 }
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ XPI_NAME := loop@mozilla.org.xpi
 XPI_FILE := $(BUILT)/$(XPI_NAME)
 
 VENV := $(BUILT)/.venv
-BABEL := $(NODE_LOCAL_BIN)/babel
+BABEL := $(NODE_LOCAL_BIN)/babel --retain-lines
 ESLINT := $(NODE_LOCAL_BIN)/eslint
 FLAKE8 := $(NODE_LOCAL_BIN)/flake8
 
@@ -243,24 +243,11 @@ $(BUILT)/add-on/install.rdf: add-on/install.rdf.in \
 # optimize this away.  We used the shared_jsx_temporary to make this code more
 # readable.
 shared_jsx_files=$(wildcard shared/js/*.jsx)
-built_ui_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/ui/%.js)
-built_add_on_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/add-on/chrome/content/%.js)
-built_standalone_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/standalone/content/%.js)
-
-# We can't use $(shared_jsx_files) here because % rules don't accept that.
-$(BUILT)/standalone/content/shared/js/%.js: shared/js/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
-
-# We can't use $(shared_jsx_files) here because % rules don't accept that.
-$(BUILT)/add-on/chrome/content/shared/js/%.js: shared/js/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
-
-# We can't use $(shared_jsx_files) here because % rules don't accept that.
-$(BUILT)/ui/shared/js/%.js: shared/js/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+shared_js_files=$(wildcard shared/js/*.js)
+shared_test_files=$(wildcard shared/test/*.js)
+built_ui_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/ui/%.js) $(shared_js_files:%.js=$(BUILT)/ui/%.js) $(shared_test_files:%.js=$(BUILT)/ui/%.js)
+built_add_on_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/add-on/chrome/content/%.js) $(shared_js_files:%.js=$(BUILT)/add-on/chrome/content/%.js) $(shared_test_files:%.js=$(BUILT)/add-on/chrome/content/%.js)
+built_standalone_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/standalone/content/%.js) $(shared_js_files:%.js=$(BUILT)/standalone/content/%.js) $(shared_test_files:%.js=$(BUILT)/standalone/content/%.js)
 
 #
 # The following sections are for the non-shared assets:
@@ -268,19 +255,15 @@ $(BUILT)/ui/shared/js/%.js: shared/js/%.jsx
 
 # ui-showcase
 ui_jsx_files=$(wildcard ui/*.jsx)
-built_ui_js_files=$(ui_jsx_files:%.jsx=$(BUILT)/%.js)
-
-$(BUILT)/ui/%.js: ui/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+ui_js_files=$(wildcard ui/*.js)
+built_ui_js_files=$(ui_jsx_files:%.jsx=$(BUILT)/%.js) $(ui_js_files:%.js=$(BUILT)/%.js)
 
 # standalone
 standalone_jsx_files=$(wildcard standalone/content/js/*.jsx)
-built_standalone_js_files=$(standalone_jsx_files:%.jsx=$(BUILT)/%.js)
-
-$(BUILT)/standalone/content/js/%.js: standalone/content/js/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+standalone_js_files=$(wildcard standalone/content/js/*.js)
+standalone_test_files=$(wildcard standalone/test/*.js)
+standalone_misc_files=$(wildcard standalone/*.js)
+built_standalone_js_files=$(standalone_jsx_files:%.jsx=$(BUILT)/%.js) $(standalone_js_files:%.js=$(BUILT)/%.js) $(standalone_test_files:%.js=$(BUILT)/%.js) $(standalone_misc_files:%.js=$(BUILT)/%.js)
 
 standalone_l10n_files=$(wildcard locale/*/standalone.properties)
 built_standalone_l10n_files=$(patsubst locale/%/standalone.properties, \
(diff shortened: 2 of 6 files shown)


.babelrc
```diff
@@ -1,6 +1,6 @@
 {
-  "plugins": [
-    "transform-react-display-name",
-    "transform-react-jsx"
+  "presets": [
+    "react",
+    "es2015"
```
Mardak, line 4: Keep this sorted

Makefile
```diff
@@ -243,44 +243,27 @@ $(BUILT)/add-on/install.rdf: add-on/install.rdf.in \
...
-
-# We can't use $(shared_jsx_files) here because % rules don't accept that.
-$(BUILT)/ui/shared/js/%.js: shared/js/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+shared_js_files=$(wildcard shared/js/*.js)
+shared_test_files=$(wildcard shared/test/*.js)
+built_ui_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/ui/%.js) $(shared_js_files:%.js=$(BUILT)/ui/%.js) $(shared_test_files:%.js=$(BUILT)/ui/%.js)
+built_add_on_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/add-on/chrome/content/%.js) $(shared_js_files:%.js=$(BUILT)/add-on/chrome/content/%.js) $(shared_test_files:%.js=$(BUILT)/add-on/chrome/content/%.js)
+built_standalone_shared_js_files=$(shared_jsx_files:%.jsx=$(BUILT)/standalone/content/%.js) $(shared_js_files:%.js=$(BUILT)/standalone/content/%.js) $(shared_test_files:%.js=$(BUILT)/standalone/content/%.js)
```
Mardak, line 250: Split these `built` variable declarations to multiline, so it's easier to parse what each substitution is doing.

Makefile
```diff
@@ -293,22 +276,29 @@ $(BUILT)/standalone/content/l10n/%/loop.properties: locale/%/standalone.properti
...
-	$(BABEL) $< --out-file $@
+	 $(add_on_jsx_files)) $(patsubst add-on/panels/js/%.js, \
+ 	 $(BUILT)/add-on/chrome/content/panels/js/%.js, \
+ 	 $(add_on_js_files)) $(patsubst add-on/panels/test/%.js, \
+ 	 $(BUILT)/add-on/chrome/content/panels/test/%.js, \
+ 	 $(add_on_test_files)) $(patsubst add-on/chrome/modules/%.jsm, \
+ 	 $(BUILT)/add-on/chrome/content/modules/%.jsm, \
+ 	 $(add_on_module_jsm_files)) $(patsubst add-on/chrome/modules/%.js, \
+ 	 $(BUILT)/add-on/chrome/content/modules/%.js, \
+ 	 $(add_on_module_js_files))
```
Mardak, line 293: The way this is split up is really hard to parse. Have each line have its own `$(patsubst .., .., ..)`

Makefile
```diff
@@ -293,22 +276,29 @@ $(BUILT)/standalone/content/l10n/%/loop.properties: locale/%/standalone.properti
...
 built_add_on_vendor_js_files=$(patsubst add-on/panels/vendor/%.jsx, \
 	 $(BUILT)/add-on/chrome/content/panels/vendor/%.js, \
-	 $(add_on_vendor_jsx_files))
-
-$(BUILT)/add-on/chrome/content/panels/vendor/%.js: add-on/panels/vendor/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+	 $(add_on_vendor_jsx_files)) $(patsubst add-on/panels/vendor/%.js, \
+	 	 $(BUILT)/add-on/chrome/content/panels/vendor/%.js, \
+	 	 $(add_on_vendor_js_files))
```
Mardak, line 301: Indention looks wrong here, but it'll be cleaned up when having `$(patsubst) on own lines anyway.

Makefile
```diff
@@ -336,11 +373,15 @@ standalone: node_modules \
             $(LODASH_OBJS) \
             $(built_standalone_js_files) \
             $(built_standalone_shared_js_files) \
-            $(built_standalone_l10n_files)
+            $(built_standalone_l10n_files) \
+						$(BUILT)/standalone/webpack.config.js \
+						$(BUILT)/standalone/content/webappEntryPoint.js
```
Mardak, line 378: Wrong indent.

Makefile
```diff
@@ -352,23 +393,26 @@ add-on: node_modules \
 	      $(BUILT)/$(ADD-ON)/chrome.manifest \
 	      $(BUILT)/$(ADD-ON)/chrome/locale/chrome.manifest \
 	      $(BUILT)/$(ADD-ON)/install.rdf \
-	      $(BUILT)/add-on/chrome/content/preferences/prefs.js
+	      $(BUILT)/add-on/chrome/content/preferences/prefs.js \
+				$(BUILT)/add-on/bootstrap.js
```
Mardak, line 397: Wrong indent.

Makefile
Mardak, line 293: Should probably just add the built `bootstrap.js` to `built_add_on_js_files` and set the appropriate `PREREQ`.

Makefile
```diff
@@ -319,15 +309,62 @@ $(BUILT)/add-on/chrome/locale/%/loop.properties: locale/%/add-on.properties loca
...
+	$(built_add_on_vendor_js_files)
+
+.SECONDEXPANSION:
+$(ALL_BUILT_JS_FILES): %: $$(PREREQ)
+	@mkdir -p $(@D)
+	$(BABEL) $< --out-file $@
+
+$(BUILT)/standalone/content/webappEntryPoint.js: standalone/content/webappEntryPoint.js
+	@mkdir -p $(@D)
+	$(BABEL) $< --out-file $@
```
Mardak, line 356: This seems to be already covered by `+$(BUILT)/standalone/content/%.js: PREREQ = standalone/content/$(@F)` ?

Makefile
Mardak, line 250: Actually, after looking more closely at what this is trying to do... we don't need all those repeated %.js matching/replacement:

``` makefile
shared_js_files=$(wildcard shared/js/*.js)
shared_jsx_files=$(wildcard shared/js/*.jsx)
shared_test_files=$(wildcard shared/test/*.js)
shared_files=$(shared_js_files) $(shared_jsx_files:.jsx=.js) $(shared_test_files)
built_add_on_shared_js_files=$(addprefix $(BUILT)/add-on/chrome/content/, $(shared_files))
built_standalone_shared_js_files=$(addprefix $(BUILT)/standalone/content/, $(shared_files))
built_ui_shared_js_files=$(addprefix $(BUILT)/ui/, $(shared_files))
```

Makefile
```diff
@@ -243,44 +243,27 @@ $(BUILT)/add-on/install.rdf: add-on/install.rdf.in \
...
 
 # ui-showcase
 ui_jsx_files=$(wildcard ui/*.jsx)
-built_ui_js_files=$(ui_jsx_files:%.jsx=$(BUILT)/%.js)
-
-$(BUILT)/ui/%.js: ui/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+ui_js_files=$(wildcard ui/*.js)
+built_ui_js_files=$(ui_jsx_files:%.jsx=$(BUILT)/%.js) $(ui_js_files:%.js=$(BUILT)/%.js)
```
Mardak, line 259: As with the shared, we only need to replace `.jsx` with `.js` for `ui_jsx_files` and just `addprefix` to both sets of files.

Makefile
Mardak, line 293: Looking more closely at each of the `patsubst`, it seems like you just want to expand `panels/` to `chrome/panels/` then `chrome/` to `chrome/content/` then prefix with `built/`?

``` makefile
add_on_js_files=$(wildcard add-on/panels/js/*.js)
add_on_jsx_files=$(wildcard add-on/panels/js/*.jsx)
add_on_module_js_files=$(wildcard add-on/chrome/modules/*.js)
add_on_module_jsm_files=$(wildcard add-on/chrome/modules/*.jsm)
add_on_test_files=$(wildcard add-on/panels/test/*.js)
add_on_files=$(add_on_js_files) $(add_on_jsx_files:.jsx=.js) $(add_on_module_js_files) $(add_on_module_jsm_files) $(add_on_test_files)
built_add_on_js_files=$(addprefix $(BUILT)/, $(subst chrome/,chrome/content/, $(subst panels/,chrome/panels/, $(add_on_files))))
```

bin/.eslintrc
```diff
@@ -1,8 +1,7 @@
 {
   "env": {
     "node": true,
-    // Can't enable this until we upgrade node.
-    "es6": false
+    "es6": true
```
Mardak, line 4: Is this change desired? I don't think we can use es6 for the `bin` js files.. ?

fzzzy, line 4: Removed this change.

.babelrc
fzzzy, line 4: Done

Makefile
fzzzy, line 259: In the shared case I agree that it is better, but in this case it doesn't seem much simpler than what's there?

Mardak, line 259: There's two places that are prefixing `$(BUILT)/` that would make it clear it's doing the same thing for all the files. Whereas with it right now, one needs to parse and recognize that each of the slightly different replacements are similar in some aspects but different in others.

fzzzy, line 259: Ok, sure.

Makefile
```diff
@@ -243,44 +243,27 @@ $(BUILT)/add-on/install.rdf: add-on/install.rdf.in \
...
 standalone_jsx_files=$(wildcard standalone/content/js/*.jsx)
-built_standalone_js_files=$(standalone_jsx_files:%.jsx=$(BUILT)/%.js)
-
-$(BUILT)/standalone/content/js/%.js: standalone/content/js/%.jsx
-	@mkdir -p $(@D)
-	$(BABEL) $< --out-file $@
+standalone_js_files=$(wildcard standalone/content/js/*.js)
+standalone_test_files=$(wildcard standalone/test/*.js)
+standalone_misc_files=$(wildcard standalone/*.js)
+built_standalone_js_files=$(standalone_jsx_files:%.jsx=$(BUILT)/%.js) $(standalone_js_files:%.js=$(BUILT)/%.js) $(standalone_test_files:%.js=$(BUILT)/%.js) $(standalone_misc_files:%.js=$(BUILT)/%.js)
```
Mardak, line 266: Just to be explicit, this can be cleaned up with the `addprefix`

Makefile
Mardak, line 301: Just to be explicit, this can be cleaned up with the `addprefix`

Makefile
Mardak, line 397: To be clear, this wrong indent doesn't matter with `bootstrap.js` as part of `built_add_on_js_files`.

PR #428, opened 2016-05-13
Bug 1272200 - Hook up the sidebar to a message manager and make it remote. Also load the ToC in a content process.

Standard8:


add-on/chrome/bootstrap.js
```diff
@@ -1226,6 +1228,9 @@ var WindowListener = {
       // remove it from existing tabs (there's no way to do that).
       window.LoopUI.mm.removeDelayedFrameScript(FRAME_SCRIPT);
 
+      let loopMessageManager = window.getGroupMessageManager("loop");
+      loopMessageManager.removeDelayedFrameScript("chrome://browser/content/content.js");
```
mancas: What if we move the loop frame script to a constant like we're doing with the general message manager `FRAME_SCRIPT`?

Standard8: Yep, I will do that. We won't need the random element for this one though, since the script is part of the browser itself.

PR #435, opened 2016-05-16
Follow-up to bug 1262441 - Add back the string that's used on the standalone infobar.

Standard8:
In reviewing & testing Chris' string detection patch, I realised that we're missing the welcome string. Although this doesn't cause an error for standalone, it does blank the infobar when we shouldn't be.

This adds the string back for now.


mozilla-autolander-deprecated:

daveccrit:
Looks good. r=dcritch

mozilla-autolander-deprecated:

PR #441, opened 2016-05-19
Bug 1274182 - Handle unexpected closing/disconnection from firebase a…

fzzzy:
…nd reconnect

### DIFF
--- a/shared/js/dataDriver.js
+++ b/shared/js/dataDriver.js
@@ -75,6 +75,8 @@ loop.DataDriver = function() {
         throw new Error("Missing option dispatcher");
       }
 
+      this._retryTimeout = this.INITIAL_RETRY_TIMEOUT;
+
       this._dispatcher = options.dispatcher;
       this._dispatcher.register(this, [
         "fetchServerData",
@@ -128,6 +130,7 @@ loop.DataDriver = function() {
       switch (event.type) {
         // Handle EventSource successfully connecting.
         case "open":
+          this._retryTimeout = this.INITIAL_RETRY_TIMEOUT;
           // XXX akita bug 1274103: Refine usage of this action.
           this._dispatcher.dispatch(new actions.DataChannelsAvailable({
             available: true
@@ -149,6 +152,27 @@ loop.DataDriver = function() {
             console.log(`Error handling EventSource.put: ${event.data}`, ex);
           }
           break;
+
+        // Handle an error from being unable to initially connect or
+        // unexpected disconnection of an open connection.
+        // Chrome follows the spec more closely than Firefox, and is
+        // able to recover from transient network conditions more reliably.
+        // Chrome will try to reconnect on error, still sending an error event,
+        // but readyState will be CONNECTING instead of CLOSED. In this
+        // case, we can just ignore the error event. Firefox sometimes sends
+        // an error event when readyState is CLOSED. At this point Firefox will
+        // not do anything to attempt to recover the connection, and thus
+        // this code runs to attempt to reconnect manually.
+        case "error":
+          if (this._sse.readyState === this._sse.CLOSED) {
+            this._sse.removeEventListener("open", this);
+            this._sse.removeEventListener("put", this);
+            this._sse.removeEventListener("error", this);
+            this._sse = null;
+            setTimeout(() => this._connectToRoom(), this._retryTimeout);
+            this._retryTimeout *= 2;
+          }
+          break;
       }
     }
 
@@ -162,6 +186,13 @@ loop.DataDriver = function() {
     get BASE_FIREBASE() {
       // XXX akita bug 1274107: Get dynamic url from loop-server.
       return "https://blinding-fire-8842.firebaseio.com/loop-test";
+    }
+
+    /**
+     * Get the initial retry delay, in ms.
+     */
+    get INITIAL_RETRY_TIMEOUT() {
+      return 2000;
     }
 
     /**
@@ -276,7 +307,7 @@ loop.DataDriver = function() {
      *
      * @param {String} token The room token to use for connections.
      */
-    _connectToRoom(token) {
+    _connectToRoom(token = this._roomToken) {
       this._roomToken = token;
       this._sse = new EventSource(this._buildUrl() + "?" + this._buildQuery({
         limitToLast: 1,
@@ -284,6 +315,7 @@ loop.DataDriver = function() {
       }));
       this._sse.addEventListener("open", this);
       this._sse.addEventListener("put", this);
+      this._sse.addEventListener("error", this);
 
       // Determine the clock skew before continuing initialization.
       this.update("meta", "lastConnect").then(({ timestamp }) => {
--- a/shared/test/dataDriver_test.js
+++ b/shared/test/dataDriver_test.js
@@ -29,7 +29,10 @@ describe("loop.DataDriver", () => {
     };
     sandbox.stub(window, "EventSource", url => {
       requests.push({ url });
-      return { addEventListener: () => {} };
+      return {
+        addEventListener() {},
+        removeEventListener() {}
+      };
     });
   });
 
@@ -148,6 +151,34 @@ describe("loop.DataDriver", () => {
           receivedTimestamp: "2009-02-13T23:31:30.123Z",
           sentTimestamp: "2016-05-10T18:26:58.235Z"
         }));
+    });
+  });
+
+  describe("#handleEvent.error", () => {
+    it("should retry the connection on error", () => {
+        // This calls _connectToRoom, so don't stub it out until after.
+        driver.setupWindowData(new actions.SetupWindowData({
+          roomToken: "fakeSetup"
+        }));
+        sandbox.stub(driver, "_connectToRoom");
+        driver.handleEvent({ type: "error" });
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT + 1);
+        sinon.assert.called(driver._connectToRoom);
+    });
+
+    it("should exponentially back off while retrying", () => {
+        driver.setupWindowData(new actions.SetupWindowData({
+          roomToken: "fakeSetup"
+        }));
+        driver.handleEvent({ type: "error" });
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT + 1);
+        expect(driver._retryTimeout).eql(driver.INITIAL_RETRY_TIMEOUT * 2);
+        driver.handleEvent({ type: "error" });
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT * 2 + 1);
+        expect(driver._retryTimeout).eql(driver.INITIAL_RETRY_TIMEOUT * 4);
+        driver.handleEvent({ type: "error" });
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT * 4 + 1);
+        expect(driver._retryTimeout).eql(driver.INITIAL_RETRY_TIMEOUT * 8);
     });
   });
 


shared/js/dataDriver.js
```diff
@@ -149,6 +151,12 @@ loop.DataDriver = function() {
             console.log(`Error handling EventSource.put: ${event.data}`, ex);
           }
           break;
+        case "error":
+          setTimeout(
+            this._connectToRoom.bind(this, this._roomToken),
```
Mardak: This seems to do the whole connection setup on error. Can we get an error after the connection has already opened? How did you test this?

If we are to use this code, it should just use `setTimeout(() => this._connectToRoom(this._roomToken))`

shared/js/dataDriver.js
```diff
@@ -277,13 +292,15 @@ loop.DataDriver = function() {
...
     _connectToRoom(token) {
+      this._retryTimeout = this.INITIAL_RETRY_TIMEOUT;
       this._roomToken = token;
       this._sse = new EventSource(this._buildUrl() + "?" + this._buildQuery({
         limitToLast: 1,
         orderBy: '"timestamp"'
       }));
       this._sse.addEventListener("open", this);
       this._sse.addEventListener("put", this);
+      this._sse.addEventListener("error", this);
```
Mardak, line 318: Does "error" happen for both disconnects and failed to initially connect?

shared/js/dataDriver.js
```diff
@@ -75,6 +75,8 @@ loop.DataDriver = function() {
         throw new Error("Missing option dispatcher");
       }
 
+      this._retryTimeout = this.INITIAL_RETRY_TIMEOUT;
```
Mardak, line 78: This seems redundant with the one inside `_connectToRoom`

shared/js/dataDriver.js
fzzzy: Yes, after the connection has been opened, the error handler will be called for a connection which is closed.

I tested by unplugging my ethernet cable.

shared/js/dataDriver.js
fzzzy, line 318: Yes. I tested by connecting, disconnecting my ethernet cable, and observing the error handler being called (with log messages). It then tries to reconnect, but can't, and the error handler is called again. The exponential backoff is operational.

shared/js/dataDriver.js
fzzzy, line 78: Yes. This is true. But it seemed weird to have a member that isn't defined until the connection is opened. Maybe I should set this to null here instead.

shared/js/dataDriver.js
Mardak: I'm not super familiar with when EventSource will try to reconnect itself, but do we run into the potential problem of having multiple connections? E.g.,

1) the original connection gave an error but is also automatically trying to reconnect
2) multiple errors can happen so we end up starting multiple recursive calls

https://html.spec.whatwg.org/multipage/comms.html#reestablish-the-connection

shared/js/dataDriver.js
Mardak, line 318: The current implementation doesn't seem like it would trigger exponential backoff as when it attempts to `_connect` again, it'll reset the delay back to initial. I'm also unclear if calling `addEventListener` with a new `_sse` but with the same callback/`this` will cause problems of receiving events from multiple `EventSource` instances.

Mardak, line 318: There's still my original question of does "error" happen for initial connection as well?

shared/js/dataDriver.js
Mardak: On a larger question.. do we actually want `_connect` to do all of what it has already done? E.g., refetching all of old chat?

shared/js/dataDriver.js
fzzzy, line 318: Whoops, you're right about the exponential backoff being reset improperly. I added the code that reset it after I tested it. Fixing.

Yes, error happens for the initial connection as well.

shared/js/dataDriver.js
fzzzy, line 78: Since the code in connectToRoom resetting this was incorrect, I added the initialization in the constructor again, and reset the timeout in the "open" event handler. I think that's the right thing to do.

shared/test/dataDriver_test.js
```diff
@@ -21,6 +21,10 @@ describe("loop.DataDriver", () => {
     driver = new loop.DataDriver({
       dispatcher
     });
+    let originalConnect = driver._connectToRoom;
+    sandbox.stub(driver, "_connectToRoom", token => {
+      originalConnect.call(driver, token);
+    });
```
Mardak: Any reason why you're stubbing this for all tests and not just for the one test? Typically you stub things so that the original function doesn't get called.

shared/test/dataDriver_test.js
```diff
@@ -151,6 +155,15 @@ describe("loop.DataDriver", () => {
     });
   });
 
+  describe("#handleEvent.error", () => {
+    it("should retry the connection on error", () => {
+        driver.handleEvent({ type: "error" });
+        sinon.assert.notCalled(dispatcher.dispatch);
```
Mardak: What action dispatch are we making sure isn't called?

shared/js/dataDriver.js
```diff
@@ -149,6 +152,12 @@ loop.DataDriver = function() {
             console.log(`Error handling EventSource.put: ${event.data}`, ex);
           }
           break;
+        case "error":
```
Mardak, line 166: Needs a newline and a comment before the `case`

shared/js/dataDriver.js
```diff
@@ -149,6 +152,12 @@ loop.DataDriver = function() {
...
+          setTimeout(
+            () => this._connectToRoom(this._roomToken),
+            this._retryTimeout);
```
Mardak: Just put these on a single line.

shared/js/dataDriver.js
```diff
@@ -149,6 +152,12 @@ loop.DataDriver = function() {
...
+          this._retryTimeout *= 2;
```
Mardak: My earlier comment still is unanswered about testing the backoff. How are you getting multiple `error` events?

shared/js/dataDriver.js
```diff
@@ -172,6 +181,13 @@ loop.DataDriver = function() {
     }
 
     /**
+     * Get the initial retry delay, in ms.
+     */
+    get INITIAL_RETRY_TIMEOUT() {
+      return 2000;
+    }
+
```
Mardak, line 204: Hmm.. I thought I pointed this out earlier, but this should be defined between `BASE_FIREBASE` and `MAX_LIMIT`.

shared/test/dataDriver_test.js
fzzzy: Just because I was looking at how the other code does it. I think it is a great idea to only stub it for one test.

shared/js/dataDriver.js
fzzzy: The error event is sent to the new EventSource created by _connectToRoom when it can't reconnect.

shared/js/dataDriver.js
fzzzy, line 204: Done

shared/test/dataDriver_test.js
```diff
@@ -151,6 +151,19 @@ describe("loop.DataDriver", () => {
...
+  describe("#handleEvent.error", () => {
+    it("should retry the connection on error", () => {
+        let originalConnect = driver._connectToRoom;
+        sandbox.stub(driver, "_connectToRoom", token => {
+          originalConnect.call(driver, token);
+        });
+        driver.handleEvent({ type: "error" });
+        sinon.assert.notCalled(dispatcher.dispatch);
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT + 1);
+        sinon.assert.called(driver._connectToRoom);
```
Mardak, line 166: It doesn't look like you're doing anything from it having called `originalConnect` so the stub can be an actual stub and not do anything.

shared/js/dataDriver.js
```diff
@@ -149,6 +152,13 @@ loop.DataDriver = function() {
             console.log(`Error handling EventSource.put: ${event.data}`, ex);
           }
           break;
+
+        // Handle an error, either unable to initially connect or
+        // unexpected disconnection
```
Mardak: nit: comments are sentences with punctuation.

shared/test/dataDriver_test.js
```diff
@@ -151,6 +151,19 @@ describe("loop.DataDriver", () => {
...
   });
 
+  describe("#handleEvent.error", () => {
+    it("should retry the connection on error", () => {
+        let originalConnect = driver._connectToRoom;
+        sandbox.stub(driver, "_connectToRoom", token => {
+          originalConnect.call(driver, token);
+        });
+        driver.handleEvent({ type: "error" });
+        sinon.assert.notCalled(dispatcher.dispatch);
```
Mardak: Repeating comment from before: What action dispatch are we making sure isn't called?

shared/js/dataDriver.js
```diff
@@ -284,6 +301,7 @@ loop.DataDriver = function() {
       }));
       this._sse.addEventListener("open", this);
       this._sse.addEventListener("put", this);
+      this._sse.addEventListener("error", this);
```
Mardak, line 318: Repeating comment from before: On a larger question.. do we actually want _connect to do all of what it has already done? E.g., refetching all of old chat?

shared/js/dataDriver.js
```diff
@@ -284,6 +301,7 @@ loop.DataDriver = function() {
       }));
       this._sse.addEventListener("open", this);
```
Mardak, line 316: It seems like we should probably check for existing `_sse` and do some cleanup as it might be possible for the old EventSource to still send events. Even if we're sure there won't be new events, we should clean up as the event handlers might keep references leaking memory.. ?

shared/js/dataDriver.js
```diff
@@ -149,6 +152,13 @@ loop.DataDriver = function() {
             console.log(`Error handling EventSource.put: ${event.data}`, ex);
           }
           break;
+
+        // Handle an error, either unable to initially connect or
+        // unexpected disconnection
+        case "error":
```
Mardak, line 166: Repeating comment from before: 

I'm not super familiar with when EventSource will try to reconnect itself, but do we run into the potential problem of having multiple connections? E.g.,

1) the original connection gave an error but is also automatically trying to reconnect
2) multiple errors can happen so we end up starting multiple recursive calls

https://html.spec.whatwg.org/multipage/comms.html#reestablish-the-connection

shared/test/dataDriver_test.js
fzzzy, line 166: You're right. When it was stubbed for all tests, it needed to call the original connect otherwise nothing would happen and other tests would fail.

shared/test/dataDriver_test.js
fzzzy: Weird, I swear I removed that.

shared/js/dataDriver.js
ianb, line 166: I guess `readyState==CONNECTING` in this reconnect case?

shared/js/dataDriver.js
ianb, line 316: Yeah, like a close method:

``` javascript
close() {
  if (this._sse) {
    this._sse.removeEventListener("open", this); // ... and other methods
  }
  this._sse = null;
}
```

Then we know there aren't shadow connections or event listeners out there that can cause confusion.

shared/js/dataDriver.js
fzzzy, line 318: The way the code is now, we only fetch one line of chat. We probably still don't want to fetch this one line if it has already been displayed. I guess the right thing to do is probably to store the most recent timestamp and fetch only stuff that has been added after that.

shared/js/dataDriver.js
fzzzy, line 166: Switched to only reconnect when readyState === CLOSED. The way the spec is written, we shouldn't ever have to manually reconnect, eventsource should take care of reconnecting in all recoverable situations. However, chrome and firefox have different behaviors that in practice means reconnecting on error when readyState === CLOSED does help for some cases in Firefox. Chrome doesn't ever need the manual code and always correctly reconnects eventually in case of network failure.

However, there are some cases when Firefox says readyState === CONNECTING, but never connects, never sends another error event, and never successfully connects. I will test for longer to see if it eventually recovers, but I'm not really sure what we can do about this case other than file a bug against gecko.

fzzzy:
Tests are failing in a strange way now.

fzzzy:
Ok, the tests are failing because EventSource is stubbed out and returns a fake object which has addEventListener but not removeEventListener. It didn't matter before I removed the event handlers on close, but now it matters.

shared/js/dataDriver.js
Mardak, line 318: What code is fetching one line of chat? on `_connect` it should be updating the time skew and gets the last 24 hours of chat.

fzzzy, line 318:
```
  this._sse = new EventSource(this._buildUrl() + "?" + this._buildQuery({
    limitToLast: 1,
```

But I see below that a GET request is also being made for the last 24 hours of chat.

shared/test/dataDriver_test.js
fzzzy, line 166: No, we do need to still call originalConnect, because otherwise this._sse is undefined.

fzzzy:
Nope, stubbing removeEventListener didn't help -- somehow this._sse is getting set to undefined, even though new EventSource is not returning undefined (which is the only place this._sse is set). _connectToServer just isn't getting called during the test for some reason.

shared/test/dataDriver_test.js
fzzzy, line 166: I found a workaround. In the test, I need to call _connectToRoom to set things up properly, but then am able to replace _connectToRoom with a do-nothing stub after things have been set up.

fzzzy:
Ok, everything seems to be in working order again here.

shared/test/dataDriver_test.js
```diff
@@ -29,7 +29,7 @@ describe("loop.DataDriver", () => {
     };
     sandbox.stub(window, "EventSource", url => {
       requests.push({ url });
-      return { addEventListener: () => {} };
+      return { addEventListener: () => {}, removeEventListener: () => {} };
```
Mardak: nit: let's split this onto multiple lines and just make them methods instead of keyed functions:

``` js
{
  addEventListener() {},
  removeEventListener() {}
}
```

shared/js/dataDriver.js
```diff
@@ -149,6 +152,18 @@ loop.DataDriver = function() {
             console.log(`Error handling EventSource.put: ${event.data}`, ex);
           }
           break;
+
+        // Handle an error from being unable to initially connect or
+        // unexpected disconnection of an open connection.
+        case "error":
+          if (this._sse.readyState === this._sse.CLOSED) {
```
Mardak, line 167: We should have a comment about the differences in Firefox and Chrome here.

shared/js/dataDriver.js
```diff
@@ -149,6 +152,18 @@ loop.DataDriver = function() {
...
+            this._sse.removeEventListener("open", this);
+            this._sse.removeEventListener("put", this);
+            this._sse.removeEventListener("error", this);
```
Mardak, line 170: Probably good to add in `this._sse = null` to break any references.

shared/js/dataDriver.js
```diff
@@ -149,6 +152,18 @@ loop.DataDriver = function() {
...
+            setTimeout(() => this._connectToRoom(this._roomToken), this._retryTimeout);
```
Mardak: Eh.. given that this is an internal API, might as well shorten this to `this._connectToRoom()` and update the definition `_connectToRoom(token = this._roomToken)` and updated doc comments for the method.

shared/test/dataDriver_test.js
```diff
@@ -151,6 +151,19 @@ describe("loop.DataDriver", () => {
...
+    it("should retry the connection on error", () => {
+        // This calls _connectToRoom, so don't stub it out until after.
+        driver.setupWindowData(new actions.SetupWindowData({
+          roomToken: "fakeSetup"
+        }));
+        sandbox.stub(driver, "_connectToRoom");
+        driver.handleEvent({ type: "error" });
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT + 1);
+        sinon.assert.called(driver._connectToRoom);
+    });
```
Mardak, line 167: Given that we've implemented backoff, we should add a second test that makes sure the backoff timer is doubled. This first test should probably check that `_retryTimer` is doubled. The second test should check that it's quadrupled after a double error.

shared/js/dataDriver.js
Mardak, line 318: As discussed on IRC, one potential downside of not always refetching is that data can change while disconnected. File a separate bug to track this refetch or not issue.

shared/test/dataDriver_test.js
fzzzy, line 167: Done.

shared/js/dataDriver.js
```diff
@@ -276,14 +307,15 @@ loop.DataDriver = function() {
      *
      * @param {String} token The room token to use for connections.
      */
-    _connectToRoom(token) {
+    _connectToRoom(token = this._roomToken) {
```
Mardak, line 310: The comment needs to be updated with the default.

shared/test/dataDriver_test.js
```diff
@@ -151,6 +154,34 @@ describe("loop.DataDriver", () => {
...
+
+    it("should exponentially back off while retrying", () => {
+        driver.setupWindowData(new actions.SetupWindowData({
+          roomToken: "fakeSetup"
+        }));
+        driver.handleEvent({ type: "error" });
+        clock.tick(driver.INITIAL_RETRY_TIMEOUT + 1);
```
Mardak, line 174: I'm not sure if we actually want to tick the clock. And it's not stubbing `_connectToRoom` so it'll actually try to connect to room 3 times. You could consolidate the `SetupWindowData` and stubbing to a `beforeEach`

PR #442, opened 2016-05-20
Bug 1273663 - make ToC open in its own window

dmose:
This probably wants a bit of cleanup and maybe some tests (which I'll do tomorrow).  That said, even in it's current form, this should unblock folks who want to base other things on the ToC stuff..

### DIFF
--- a/add-on/chrome/modules/LoopRooms.jsm
+++ b/add-on/chrome/modules/LoopRooms.jsm
@@ -3,7 +3,7 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-const { utils: Cu } = Components;
+const { classes: Cc, interfaces: Ci, utils: Cu } = Components;
 
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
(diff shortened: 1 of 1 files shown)


add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -746,9 +746,26 @@ var LoopRoomsInternal = {
...
+    let windowArgs =
+      Cc["@mozilla.org/supports-array;1"].createInstance(Ci.nsISupportsArray);
+    windowArgs.AppendElement(urlToLoad);
+
+    // XXX Do we need to do anything more sophisticated (eg handle private
+    // browsing mode)?  If so, we may want to use utilityOverlay.js:openUILinkIn
+    // (which calls utilityOverlay.js:openLinkIn).  Doing this is rather gross,
+    // as we would need to get the most recently used window and then reach in
+    // and invoke window.openUILinkIn, and it's also conceivably fragile, since
+    // this is unlikely to be a supported way to use that API.
```
Standard8: Digging through `openLinkIn`, it actually ends up calling `Services.ww.openWindow` as you do here, but with `",private"` on the end of the features.

When I then follow that `OpenWindow` through into nsWindowWatcher.cpp it appears that all the "features" are hard-coded in c++ land. That's going to make it super difficult for us to have a specific type of window.

Since for various things we need to be able to detect if a window is our own or not, what we could do is when a window is opened, and `WindowListener#onOpenWindow` (in bootstrap) detects it, then we can check the url being loaded, if its `about:looptoc` then we can set a flag on the window, we can also potentially use that to add our extra UI.

We don't have to do that all here, but I'm thinking that may be a better thing to start to work towards.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -746,9 +746,26 @@ var LoopRoomsInternal = {
...
+
+    // XXX is there any reason to bother passing in the real parent?
```
Standard8: I can't think of one.

add-on/chrome/modules/LoopRooms.jsm
dmose: @Standard8: I assume "set a flag", you mean "set an HTML attribute", since we'll be able to use that in CSS classes.  Or did you have something else in mind?

Standard8: I was more thinking something that could be easily accessed by `window.foo` (similar to `window.LoopUI` but specific to a room window, maybe with the room token?

dmose: As I mentioned in the bug, I'd vote for doing this later.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -746,9 +746,26 @@ var LoopRoomsInternal = {
...
+      Cc["@mozilla.org/supports-array;1"].createInstance(Ci.nsISupportsArray);
+    windowArgs.AppendElement(urlToLoad);
+
+    // If we ever decide we want to do anything slightly more sophisticated (eg
+    // allow Hello in private browsing mode), we may want to use
+    // utilityOverlay.js:openUILinkIn (which calls
+    // utilityOverlay.js:openLinkIn).  Doing this is rather gross, as we would
+    // need to get the most recently used window and then reach in and invoke
+    // window.openUILinkIn, and it's also conceivably fragile, since this is
+    // unlikely to be a supported way to use that API.
```
Standard8, line 763: I think I'd like to see the alternative that I suggested added here as well. Something like "Alternately, we set an object and token on the window object for the newly created window, and use that as our flag".

PR #454, opened 2016-05-23
Bug 1273495 - Create Tile to display the added URL information

mancas:


mozilla-autolander-deprecated:

shared/js/toc.jsx
```diff
@@ -303,26 +303,42 @@ loop.shared.toc = (function(mozL10n) {
       page: React.PropTypes.object.isRequired
     },
 
+    deleteTile: function() {
+      // XXX delete tile from firebase
+      console.info("Tile deleted");
```
fcampo: Till the bug is fixed, change it to `console.info("Tile should be deleted from Firebase at this point");`

PR #458, opened 2016-05-24
Bug 1272655 - Move the sidebar code into its own module

mancas:


add-on/chrome/modules/LoopSidebar.jsm
```diff
@@ -0,0 +1,85 @@
...
+    // message manager for the sidebar.
+    let loopMessageManager = window.getGroupMessageManager("loop");
+    loopMessageManager.loadFrameScript(BROWSER_FRAME_SCRIPT, true);
+  },
+
+  createSidebar: function(gBrowser) {
+    LoopSidebarInternal.createSidebar(gBrowser);
+  },
+
+  loadSidebar: function(token) {
```
Standard8: Having looked at this a bit, I think we could pretty easily combine `createSidebar` and `loadSidebar`. There's no reason to create one without having loaded it first.

Also, we could move the `loadFrameScript` code from init() into the `loadSidebar` function - we don't need that frame script being able to be loaded until just before we have a sidebar.

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,20 @@
...
+
+const { LoopSidebar } = Cu.import("chrome://loop/content/modules/LoopSidebar.jsm", {});
+
+add_task(function* test_LoopSidebar_noSidebar() {
+  let sidebar = window.document.getElementById("loop-side-browser");
+  Assert.ok(!sidebar, "there should be no sidebars yet");
+});
+
+add_task(function* test_LoopSidebar_createSidebar() {
+  LoopSidebar.createSidebar(window);
```
Standard8: I'd suggest changing these tests to create the sidebar on a new window, and then close the window afterwards. That way we have less chance of causing side-effects on other tests.

There's functions in BrowserTestUtils.jsm that will help with this in e10s compatible ways.

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,20 @@
...
+  let sidebar = window.document.getElementById("loop-side-browser");
+  Assert.ok(sidebar, "there should create a sidebar");
```
Standard8, line 46: We should be able to load the sidebar as well, and check that the `src` is `about:loopconversation#faketoken`

Standard8, line 46: I forgot to mention, when loading the sidebar, we should go into offline mode to avoid any potential network activity (the tests will complain), `loadLoopPanel` in head.js has an example of how to do this.

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,20 @@
...
+});
```
Standard8, line 49: Once we've loaded the sidebar, we should be able to check the permissions as well, something like:

`Assert.equal(Services.perms.testPermission(uri, "camera"), Services.perms.ALLOW_ACTION, "failure message")`

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,52 @@
...
+  Assert.strictEqual(sidebar.getAttribute("src"), "about:loopconversation#faketoken", "Url must be set properly");
+});
+
+add_task(function* test_LoopSidebar_checkPermissions() {
+  goOffline();
+  win = yield BrowserTestUtils.openNewBrowserWindow();
+  LoopSidebar.createSidebar(win, "faketoken");
+  Assert.equal(Services.perms.testPermission("about:loopconversation#faketoken", "camera"),
+    Services.perms.ALLOW_ACTION, "Camera permissions should be granted");
+});
```
mancas, line 59: @Standard8 I'm getting an error when running the last test:

> 25 INFO TEST-UNEXPECTED-FAIL | browser/extensions/loop/chrome/test/mochitest/browser_LoopSidebar.js | Uncaught exception - [Exception... "Could not convert JavaScript argument arg 0 [nsIPermissionManager.testPermission]"  nsresult: "0x80570009 (NS_ERROR_XPC_BAD_CONVERT_JS)"  location: "JS frame :: chrome://mochitests/content/browser/browser/extensions/loop/chrome/test/mochitest/browser_LoopSidebar.js :: test_LoopSidebar_checkPermissions :: line 49"  data: no`

It seems that complains about the result of testPermission but according to mdn documentation, that function returns a boolean

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,65 @@
...
+
+add_task(function* test_LoopSidebar_createSidebar() {
+  goOffline();
+  win = yield BrowserTestUtils.openNewBrowserWindow();
+  LoopSidebar.createSidebar(win, "faketoken");
+  let sidebar = win.document.getElementById("loop-side-browser");
+  Assert.ok(sidebar, "there should create a sidebar");
+  Assert.strictEqual(sidebar.getAttribute("src"), "about:loopconversation#faketoken", "Url must be set properly");
+  closeWindow();
```
Standard8: `closeWindow` is a generator function, so you need to `yield` to it.

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,65 @@
...
+
+  registerCleanupFunction(function() {
+    Services.io.offline = WAS_OFFLINE;
+  });
+}
+
+add_task(function* test_LoopSidebar_noSidebar() {
+  let sidebar = window.document.getElementById("loop-side-browser");
+  Assert.ok(!sidebar, "there should be no sidebars yet");
+  closeWindow();
```
Standard8: No need for this `closeWindow` here (you're not opening one)

add-on/chrome/test/mochitest/browser_LoopSidebar.js
```diff
@@ -0,0 +1,65 @@
...
+
+registerCleanupFunction(function* () {
+  if (win) {
+    yield BrowserTestUtils.closeWindow(win);
+  }
+});
+
+function* closeWindow() {
+  if (!win) {
+    win = window;
```
Standard8: We shouldn't need the if() here, nor the assignment, if its null, there's an error.

add-on/chrome/bootstrap.js
```diff
@@ -1230,8 +1193,7 @@ var WindowListener = {
       // remove it from existing tabs (there's no way to do that).
       window.LoopUI.mm.removeDelayedFrameScript(FRAME_SCRIPT);
 
-      let loopMessageManager = window.getGroupMessageManager("loop");
-      loopMessageManager.removeDelayedFrameScript(BROWSER_FRAME_SCRIPT);
+      LoopSidebar.tearDown();
```
Standard8: You need to pass in `window` here.

add-on/chrome/modules/LoopRooms.jsm
```diff
@@ -33,6 +33,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "loopCrypto",
   "chrome://loop/content/shared/js/crypto.js", "LoopCrypto");
 XPCOMUtils.defineLazyModuleGetter(this, "ObjectUtils",
   "resource://gre/modules/ObjectUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "LoopSidebar",
+  "chrome://loop/content/modules/LoopSidebar.jsm");
```
Standard8: I don't think you need this here any more.

add-on/chrome/modules/LoopSidebar.jsm
```diff
@@ -0,0 +1,78 @@
...
+  let ConsoleAPI = Cu.import("resource://gre/modules/Console.jsm", {}).ConsoleAPI;
+  let consoleOptions = {
+    maxLogLevelPref: PREF_LOG_LEVEL,
+    prefix: "Loop"
+  };
+  return new ConsoleAPI(consoleOptions);
+});
+
+var LoopSidebarInternal = {
+  createSidebar: function(window, token) {
```
Standard8, line 39: nit: please add jsdoc for the functions in the file.

PR #464, opened 2016-05-26
Bug 1274168 - Store participant data in firebase

Mardak:


shared/js/actions.js
```diff
@@ -126,6 +126,14 @@ loop.shared.actions = (function() {
     }),
 
     /**
+     * Notifies that participant data has been received.
+     */
+    UpdatedParticipant: Action.define("updatedParticipant", {
+      name: String,
+      userId: String
```
dmose, line 134: It'd be helpful to put the format/origin of the userId in the comments, even if we're currently just doing something hacky for bootstrapping.

shared/js/dataDriver.js
```diff
@@ -305,15 +316,27 @@ loop.DataDriver = function() {
...
       switch (type) {
-        case "chat": {
-          let chat = data.value;
-          chat.receivedTimestamp = new Date(data.timestamp).toISOString();
-          this._dispatcher.dispatch(new actions.ReceivedTextChatMessage(chat));
+        case "chat":
+          dispatchAction = "ReceivedTextChatMessage";
+          dispatchObject = Object.assign(data.value, {
+            receivedTimestamp: new Date(data.timestamp).toISOString()
+          });
```
dmose: I'm not a huge fan of writing to formal parameters -- it's error prone, since JavaScript passes a copy of object references.  In this case, I think it means that the changes being written into data.value for local use will also be visible to the caller after this returns.  If that's actually how this is behaving, I'd suggest changing the behavior to not write data.value.  If that's important for some reason, however, at the very least it wants to be documented in the JSDoc that this method has intentional side effects.

dmose:
r=@dmose with comments addressed

shared/js/actions.js
Mardak, line 134: Added comment saying it's from loop-server.

shared/js/dataDriver.js
Mardak: Avoided modifying `data` param by refactoring to have a `dispatchExtra` object and combined them with `.assign` that starts with `{}` then adds `data.value` then adds `dispatchExtra`.

Mardak: Also added unit tests making sure params aren't modified.

dmose: Awesome; thanks!  There's actually an eslint rule we should probably consider turning on for this: 
http://eslint.org/docs/rules/no-param-reassign

PR #466, opened 2016-05-27
Bug 1274181 - Request existing room data after firebase connects

Mardak:
This required fixing loop_mocha_utils to correctly implement Promises as SyncThenable.

### DIFF
--- a/shared/js/dataDriver.js
+++ b/shared/js/dataDriver.js
@@ -275,6 +275,23 @@ loop.DataDriver = function() {
     }
 
     /**
+     * Request all room data except for chat.
+     *
+     * @return {Promise} Resolved with an {Object} with keys of record ids.
+     */
+    requestRoomData() {
+      let query = this._buildQuery({
+        orderBy: '"$key"',
+        // NB: All record types must come lexicographically after "chat" for
+        // this query to request all records except for type "chat".
+        startAt: '"chat~"',
+        // NB: Firebase requires a limit to have things sorted correctly.
+        limitToLast: this.MAX_LIMIT
+      });
+      return this._request("GET", this._buildUrl(), query);
+    }
+
+    /**
      * Update a record additionally storing the timestamp.
      *
      * @param {String} type  Type of record to update.
@@ -326,9 +343,11 @@ loop.DataDriver = function() {
     }
 
     /**
-     * Connect to a specified room for streaming updates and future requests.
+     * Connect to a specified room for streaming updates and future requests
+     * then initialize with existing data.
      *
-     * @param {String} token The room token to use for connections.
+     * @param {String} token The room token to use for connections defaulting to
+     *                       the current room for reconnect.
      */
     _connectToRoom(token = this._roomToken) {
       this._roomToken = token;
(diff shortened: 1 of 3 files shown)


shared/test/loop_mocha_utils.js
```diff
@@ -43,8 +43,18 @@ var LoopMochaUtils = (function(global, _) {
...
+        continuations.push(result => {
+          let ret = contFn(result);
+          if (ret instanceof SyncThenable) {
+            ret.then(resolve);
+          }
+          else {
+            resolve(ret);
+          }
+        });
+      });
```
dmose, line 60: It'd be nice to make this a bit more readable on first approach.  It might help to make the outer arrow function a separate function, though I kinda suspect not.  Maybe just another comment or two as to how the mechanics work?

shared/test/loop_mocha_utils.js
```diff
@@ -80,14 +90,18 @@ var LoopMochaUtils = (function(global, _) {
   SyncThenable.all = function(promises) {
     return new SyncThenable(function(resolve) {
       var results = [];
+      let resultCount = 0;
 
-      promises.forEach(function(promise) {
+      promises.forEach(function(promise, i) {
         promise.then(function(result) {
-          results.push(result);
+          results[i] = result;
```
dmose, line 101: It might be worth a comment that the index from the iterable is used intentionally to maintain the semantic of Promise.all that guarantees the results are in the order passed in rather than the order resolved, just so that if someone needs to change this code in the future, they don't inadvertently break that.

shared/js/dataDriver.js
```diff
@@ -343,11 +359,17 @@ loop.DataDriver = function() {
       // Determine the clock skew before continuing initialization.
       this.update("meta", "lastConnect").then(({ timestamp }) => {
```
dmose, line 363: I'd suggest that the arrow function move into it's own top-level function, since it actually does something other than just connecting to the room.  I would expect it to also simplify your unit tests, since there would be two simple units in play rather than one more complex one.

If you feel strongly that this won't be the worth the work, you can skip it.  

That said, in either case, it'd be great if _connectToRoom were _connectToRoomAndUpdate or something similar so that folks reading the calling code understand the implied side-effects to some degree.

dmose, line 363: Or, alternately, if you think the splitting makes sense, but isn't worth doing now, I'd suggest updating the JSDoc and probably putting an XXX in the code to that effect.

shared/test/dataDriver_test.js
```diff
@@ -298,6 +298,17 @@ describe("loop.DataDriver", () => {
...
   });
 
+  describe("#requestRoomData", () => {
+    it("should request everything but chat", () => {
+      driver.requestRoomData();
+
+      let queries = requests[0].url.split("?")[1].split("&").sort();
+      expect(queries).contain("orderBy=%22%24key%22");
+      expect(queries).contain("startAt=%22chat~%22");
+      expect(queries).contain(`limitToLast=${driver.MAX_LIMIT}`);
```
dmose: This wants a few helpers, probably on dataDriver itself, to make this more readable.  That said, feel free to just add an XXX here so that someone considers doing it in the future.

shared/js/dataDriver.js
```diff
@@ -275,6 +275,22 @@ loop.DataDriver = function() {
...
 
     /**
+     * Request all room data except for chat.
+     *
+     * @return {Promise} Resolved with an {Object} with keys of record ids.
+     */
+    requestRoomData() {
+      let query = this._buildQuery({
+        orderBy: '"$key"',
+        // NB: All record types come lexicographically after chat.
```
dmose: Is this an invariant that we need to maintain when adding new record types?  If so, how can we ensure that that happens?

dmose:
Looks good; r=dmose with comments addressed as appropriate.

shared/test/loop_mocha_utils.js
Mardak, line 101: Commented.

shared/test/loop_mocha_utils.js
Mardak, line 60: Commented.

shared/test/dataDriver_test.js
Mardak: I added some test helpers to call the existing `dataDriver._buildQuery` that does the url encoding.

shared/js/dataDriver.js
Mardak, line 363: I split out the code starting from Promise.all as the "update" for lastConnect and setting `_clockSkew` really are part of connecting. The update is just how dataDriver reads out the server time, but it doesn't actually update room data.

shared/js/dataDriver.js
Mardak: Yup. Left a comment in https://bugzilla.mozilla.org/show_bug.cgi?id=1274110#c2

dmose: Cool.  I'd still suggest making the comment here more explicit.

Mardak: Expanded the comment.

PR #467, opened 2016-05-27
Bug 1274137 - (akita-participants-store) stand up basic ParticipantStore. r=Mardak

mancas:


shared/js/participantStore.js
```diff
@@ -0,0 +1,104 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var loop = loop || {};
+loop.store = loop.store || {};
+
+loop.store.ParticipantStore = (function() {
+  "use strict";
+
+  const PARTICIPANT_SCHEMA = {
+    name: "",
```
Mardak, line 12: Does having `name` work correctly? I think `Actions` actually have a `name` of the action itself, e.g., `updatedParticipant` and `updatedPresence`.

shared/js/participantStore.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      return updatedParticipant;
+    },
+
+    _getOnlineParticipants: function() {
+      let participants = this._storeState.participants;
+      let onlineParticipants = [];
+      participants.forEach(function(participant) {
+        if (participant.isHere) {
```
Mardak: We'll need a check for pingTime for any unexpected disconnects/leaves. I'm not sure what the timeout should be as we currently don't have continuous presence updates yet.

shared/js/participantStore.js
```diff
@@ -0,0 +1,104 @@
...
+    pingTime: null
+  };
+
+  /**
+   * Participant store.
+   *
+   * @param {loop.Dispatcher} dispatcher  The dispatcher for dispatching actions
+   *                                      and registering to consume actions.
+   * @param {Object} options Options object:
+   * - {OTSdkDriver} sdkDriver  The SDK driver instance.
```
Mardak: It doesn't look like `OTSdkDriver` is neccessary.

shared/js/participantStore.js
```diff
@@ -0,0 +1,104 @@
...
+    updatedPresence: function(actionData) {
+      let participant = this._updateParticipantData(actionData.userId, {
+        isHere: actionData.isHere,
+        pingTime: actionData.pingTime
+      });
+
+      let updatedParticipants = this._storeState.participants.set(actionData.userId, participant);
+
+      this.setStoreState({
+        participants: updatedParticipants
```
Mardak: This being slightly different from the pattern in `updatedParticipant` is somewhat confusing. Might as well just move this logic into `_updateParticipantData`.

shared/js/participantStore.js
mancas, line 12: @Mardak It works because the property is overwritten. I don't like it so, what if I change the action to:
`UpdatedParticipant: Action.define("updatedParticipant", {
      participantName: String...
}),`

do you agree?

Mardak, line 12: Sounds good. Thanks!

mancas:
@Mardak can you take a quick look at the pingtime timeout?

shared/js/participantStore.js
```diff
@@ -0,0 +1,96 @@
...
+    getOnlineParticipants: function() {
+      let participants = this._storeState.participants;
+      let onlineParticipants = [];
+      participants.forEach(function(participant) {
+        let pingTime = new Date(participant.pingTime);
+        let maxPingTimeAllowed = new Date();
+        // XXX akita: Max ping time allowed is not defined yet (1 minute by default)
+        // If you change the timeout, don't forget to update the test
+        maxPingTimeAllowed.setMinutes(maxPingTimeAllowed.getMinutes() - 1);
+        if (participant.isHere && pingTime.getTime() >= maxPingTimeAllowed.getTime()) {
```
Mardak: Doh sorry, this won't quite work. We'll need to change the `UpdatedPresence` action and `DataDriver` class. This is broken because it's comparing server time to local time. Instead `DataDriver` should dispatch an object containing something like.. `pingedAgo` where locally it can do `this.getServerTime() - data.timestamp` so then this `getOnlineParticipants` code just checks if the "ago" value is greater than 1 minute (60000 milliseconds).

mancas: @Mardak hmm not sure if it's going to work as expected. AFAIK the `pingedAgo` is calculated everytime the user send a "presence event" but let's imagine that there is an unexpected disconnection then the latest pingedAgo is always correct and is not greater than 1minute so the user is "online". What I mean is that we need to compere the pingedAgo value with the current time. Does it makes sense or am I wrong?

Mardak: Ah right. The ago/delta value will stay constant when stored in the `Map`. I suppose we could additionally store when the presence action was received so we could do something like... `pingedAgo - receivedPresenceTime + now > 1 minute`.

Mardak: To be clear regarding:

```
imagine that there is an unexpected disconnection then the latest pingedAgo is always correct and is not greater than 1minute so the user is "online"
```

`pingedAgo` would be calculated from the current server time, so the ago value would be pretty accurate on first load of the page. However, subsequent calls to `getOnlineParticipants` (e.g., a minute later to refresh the presence indicators) would compare with stale values if not updated with the current time.

Mardak: I think it makes sense to have the UpdatedParticipant action to have `pingedAgo` instead of `pingTime` to avoid needing to expose action handlers to `DataDriver` to get the server time. Within `ParticipantStore` on receiving a `pingedAgo`, the value stored into the `Map` could be a single value of `localPingTime: Date.now() - pingedAgo` then in the dynamic calls to `getOnlineParticipants` calculate `Date.now() - localPingTime > 1 minute`

mancas: I see, that could work. However, I've been talking with Fernando, and we have a question about Firebase. Seems that having a timeout or a similar mechanism to update the view by calling to `getOnlineParticipants` is quite weird since React views should be updated by changes in the store.

So, there is any way to track the user presence in Firebase? From out point of view it should works like this: If Firebase detects that an user hasn't sent a ping in the latest minute then a disconnect event is sent to the dataDriver.

mancas: Doing a quick search in google I found these articles:

https://firebase.googleblog.com/2013/06/how-to-build-presence-system.html

https://groups.google.com/forum/#!msg/firebase-talk/fjzc00gxsQU/-Xd-VFEWn0MJ

It seems that the onDisconnect event is thrown when Firebase detects that a socket has been closed. In that case, if we have a, unexpected crash, the socket will be closed and Firebase can send that event. Am I missing something?

ianb: I looked into onDisconnect some, and it didn't seem available through the REST/EventSource API, only with Firebase.js

Mardak: As ianb noted, we're using REST APIs, so there isn't an automatic onDisconnect. Even if we implemented something similar, an unclean shutdown of Firefox, e.g., app/system crash or power outage would have these dangling presence records.

If we want the store to be modified, then the suggestion for calculating `localPingTime` should work with `ParticipantStore` peridically refreshing that value for each participant.

shared/js/participantStore.js
```diff
@@ -0,0 +1,96 @@
...
+      this.setStoreState({
+        participants: this._storeState.participants.set(userId, updatedParticipant)
+      });
+    },
+
+    /*
+     * Gets the online participants in the room taking into account
+     * unexpected disconnects/leaves
+     */
+    getOnlineParticipants: function() {
```
dmose, line 76: I'm a little confused by this API.  If the code for getting the subset of participants who are online is worth centralizing, which I suspect it is, I'd expect the API to it to be some data in the store state, so that views will automatically update as the store state changes.  Is there a reason to make the API be a function that I'm missing?

Mardak, line 76: The store state does change via `_updateParticipantData` that calls `this.setStoreState`. Is there a particular benefit of accessing plain objects instead of the `Map` and helper?

dmose, line 76: You're right that it would work.  It just seems a bit out-of-pattern to have to call back into the store for a helper, rather than just having the data available in this.state in processed form.  But I suppose we can try it and change it if there's a problem.

mancas:
@Mardak Patch ready for the last "review". I've addressed your comments and updated unit tests

shared/js/participantStore.js
mancas, line 76: I agree that having that helper function seems a bit out-of-pattern, however, if we don't have the filter logic in the store we'll need to "copy&paste" it everytime we want to use the presence in the UI. But you're right @dmose let's try this solution and let's decide later if we need to refactor it or not.

dmose, line 76: Yeah, totally good enough for now.

If we do decide to refactor, what I was unsuccessfully trying to communicate was that a way to do that would be to leave that code in the store, but to somehow change the data store to expose the pre-filtered data.  I could imagine a couple of ways of doing it at least (splitting the data into active and inactive objects, or denormalizing and having an unfiltered and filtered object).

shared/js/participantStore.js
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Checks if the room is empty or has participants.
+     */
+    _hasParticipants: function() {
+      return this._storeState.participants.size > 0;
+    },
+
+    updatedParticipant: function(actionData) {
+      this._updateParticipantData(actionData.userId, {
+        name: actionData.name
```
Mardak, line 44: This should be reading actionData.participantName now

shared/test/participantStore_test.js
```diff
@@ -0,0 +1,139 @@
...
+      expect(initialState.participants).not.eql(null);
+    });
+  });
+
+  describe("updatedParticipant", function() {
+    it("should store the participant", function() {
+      let action = {
+        userId: "fakeID",
+        name: "fakeName"
+      };
```
Mardak, line 42: Probably should wrap this in the actual action constructor, so it would have caught the misnamed name -> participantName.

shared/test/dataDriver_test.js
```diff
@@ -506,23 +506,24 @@ describe("loop.DataDriver", () => {
...
         }));
     });
   });
 
   describe("#_processRecord.presence", () => {
     it("should dispatch `UpdatedPresence` for presence record", () => {
+      clock.tick(1234567890123);
       driver._processRecord("presence!theUserId", {
-        timestamp: 1234567890123,
+        timestamp: 1234567889123,
```
Mardak, line 526: This should be a computed number as it's easy to look over the difference between this and the tick amount above.

shared/js/participantStore.js
```diff
@@ -0,0 +1,93 @@
...
+      let onlineParticipants = [];
+      participants.forEach(function(participant) {
+        // XXX akita: Max ping time allowed is not defined yet (1 minute by default)
+        // If you change the timeout, don't forget to update the test
+        if (participant.isHere && Date.now() - participant.localPingTime <= 60000) {
+          onlineParticipants.push(participant);
+        }
+      });
+
+      return onlineParticipants;
```
Mardak: This can all be simplified to a `[].filter`.

Mardak:
Superseded by #479.

PR #469, opened 2016-05-30
Bug 1274130 - Validate data when processing firebase records

mancas:


shared/js/dataDriver.js
```diff
@@ -437,5 +462,25 @@ loop.DataDriver = function() {
         request.send(JSON.stringify(body) || null);
       });
     }
+
+    _validateData(type, data) {
+      let schema = this._getSchemaByType(type);
+      let validatedData = new loop.validate.Validator(schema).validate(data || {});
```
Mardak: We should always have data, so if it's not there, we should probably fail.

shared/js/dataDriver.js
```diff
@@ -437,5 +462,25 @@ loop.DataDriver = function() {
...
+
+      return validatedData;
```
Mardak: Did you mean to do something with this return value?

shared/js/dataDriver.js
```diff
@@ -437,5 +462,25 @@ loop.DataDriver = function() {
...
+    _getSchemaByType(type) {
+      switch (type) {
+        case "chat":
+          return CHAT_VALIDATOR_SCHEMA;
+        case "participant":
+          return PARTICIPANT_VALIDATOR_SCHEMA;
+        case "presence":
+          return PRESENCE_VALIDATOR_SCHEMA;
+        default:
+          return {};
```
Mardak: An unexpected type should probably make the record invalid.

shared/js/dataDriver.js
```diff
@@ -7,6 +7,22 @@ loop.DataDriver = function() {
...
+  /*
+   * When creating new schemas, be sure to add them to _getSchemaByType method
+   */
+  const CHAT_VALIDATOR_SCHEMA = {
+    timestamp: Number,
+    value: Object
+  };
+
+  const PARTICIPANT_VALIDATOR_SCHEMA = {
+    value: Object
```
Mardak: Any reason no timestamp here? All schemas have to be one level deep for Validator, and all records should have at most two fields of timestamp and optionally value (for deleted records).

shared/test/dataDriver_test.js
```diff
@@ -455,9 +459,35 @@ describe("loop.DataDriver", () => {
...
+    });
+
+    it("should throw an exception if value isn't valid", () => {
+      driver._processRecord("chat!01234567abcdefghijkl", {
+        timestamp: 1234567890123,
+        value: 12412
+      });
+
+      sinon.assert.calledOnce(console.error);
+    });
```
Mardak, line 527: Add a test for `timestamp: NaN` where `typeof NaN` is "number".

shared/js/dataDriver.js
```diff
@@ -360,10 +376,19 @@ loop.DataDriver = function() {
      *  - {Mixed}  value     Stored value for the record.
      */
     _processRecord(key, data) {
-      let dispatchAction;
-      let dispatchExtra = {};
       // XXX akita bug 1274130: Validate data/key/timestamp before processing.
       let [, type, id] = key.match(/^([^!]+)!(.+)$/);
```
Mardak: The `key` should be validated although you could probably just do this by saving the result of `match` to check if it's `null` before doing the destructured assignment.

shared/js/dataDriver.js
```diff
@@ -360,10 +376,19 @@ loop.DataDriver = function() {
      *  - {Mixed}  value     Stored value for the record.
      */
     _processRecord(key, data) {
-      let dispatchAction;
-      let dispatchExtra = {};
       // XXX akita bug 1274130: Validate data/key/timestamp before processing.
```
Mardak: Don't forget to delete this XXX comment when we're done with this bug! :)

shared/js/dataDriver.js
Mardak: Alternatively declare `let type, id;` outside of the try and just do the destructured assignment on the `match` where if there's no match, there'll be a `TypeError: key.match(...) is null`

shared/js/dataDriver.js
```diff
@@ -360,10 +376,19 @@ loop.DataDriver = function() {
...
       let [, type, id] = key.match(/^([^!]+)!(.+)$/);
+      try {
+        // First validates data, if no exception is thrown we can dispatch the actions
+        this._validateData(type, data);
+      } catch (exception) {
+        console.error("Validate data error: ", exception);
```
Mardak: We're technically validating not just "data" but the "record" which has the key (including type) and the data (including timestamp and value). ;)

shared/test/dataDriver_test.js
```diff
@@ -473,9 +503,22 @@ describe("loop.DataDriver", () => {
           userId: "theUserId"
         }));
     });
+
+    it("should throw an exception if value isn't valid", () => {
+      driver._processRecord("participant!theUserId", {
+        timestamp: 1234567890123,
+        value: 1234
```
Mardak, line 567: Technically the value can be of any value. E.g., if we had a `meta!roomName` record, the value would probably be a string and not an object.

Edit: Oh, but I see this is specifically for the participant.

Standard8:

PR #471, opened 2016-05-31
Bug 1273512 - Show room"s other participants presence when connected

mancas:


shared/js/toc.jsx
```diff
@@ -125,27 +128,32 @@ loop.shared.toc = (function(mozL10n) {
...
-            <span>{'P'}</span>
-          </div>
-          <div className="room-user" data-name="Manu">
-            <span>{'M'}</span>
-          </div>
+          {
+            this.props.participantStore.getOnlineParticipants().map(function(participant, index) {
+              return (
+                <div className="room-user" data-name={participant[1].participantName} key={index}>
+                  <span>{participant[1].participantName.substring(0, 1).toUpperCase()}</span>
```
Mardak: Doh. Could you fix up getOnlineParticipants to just return an array of participants instead of array of `[key, value]`? Should be able to do `[...this._storeState.participants.values()]` and update the method documentation with the return value type and update test to check that the online participants returned is a participant with a name?

The lines here would then just be `participant.participantName`

shared/test/toc_test.js
```diff
@@ -0,0 +1,65 @@
...
+      let activeParticipants = ReactDOM.findDOMNode(view).querySelectorAll(".room-user");
+      expect(activeParticipants.length).eql(1);
+    });
+
+    it("should no render bubbles if there is no online users", () => {
+      participant.isHere = false;
+      mountTestComponent();
+      let activeParticipants = ReactDOM.findDOMNode(view).querySelectorAll(".room-user");
+      expect(activeParticipants.length).eql(0);
+    });
```
Mardak, line 73: Probably should test the 2 online participant case as that's the main use case.

shared/test/toc_test.js
```diff
@@ -0,0 +1,65 @@
...
+  describe("ToC.RoomPresenceView", () => {
+    let view, participant;
+
+    function mountTestComponent() {
+      let participants = new Map();
+      participants.set("fakeID", participant);
+
+      participantStore.setStoreState({
+        participants
+      });
```
Mardak, line 43: We should probably use the ParticipantStore APIs of `updatedParticipant` and `updatedPresence` to avoid reaching into the internals of a separate class.

shared/test/toc_test.js
```diff
@@ -0,0 +1,65 @@
...
+        participantName: "fakeName",
+        localPingTime: Date.now(),
+        isHere: true
+      };
+    });
+
+    it("should render the participant bubble", () => {
+      mountTestComponent();
+      let activeParticipants = ReactDOM.findDOMNode(view).querySelectorAll(".room-user");
+      expect(activeParticipants.length).eql(1);
```
Mardak, line 62: We should make sure the correct content is rendered: "F"

shared/test/toc_test.js
```diff
@@ -0,0 +1,65 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+describe("loop.TableOfContents", () => {
+  "use strict";
+
+  let { expect } = chai;
+  let TestUtils = React.addons.TestUtils;
```
Mardak: nit: `let { TestUtils } = React.addons;`

shared/js/toc.jsx
Mardak: Also, there's a simpler way to get the first character of a string ;) `participantName[0]`

mancas: @Mardak  I think the timestaps calculated in the store are wrong. Let's imagine I've updated my ping with Date.now() - 1000 so the localPingTime=1000 then I invoke getOnlineParticipants and Date.now() - participant.localPingTime <= 60000 is false because of that, I'm not online

shared/js/participantStore.js
```diff
@@ -41,14 +41,14 @@ loop.store.ParticipantStore = function() {
...
-        name: actionData.participantName
+        participantName: actionData.participantName
       });
     },
 
     updatedPresence(actionData) {
       this._updateParticipantData(actionData.userId, {
         isHere: actionData.isHere,
-        localPingTime: Date.now() - actionData.pingedAgo
+        localPingTime: actionData.pingedAgo
```
mancas: @Mardak I think the timestaps calculated in the store are wrong. Let's imagine I've updated my ping with Date.now() - 1000 so the localPingTime=1000 then I invoke getOnlineParticipants and Date.now() - participant.localPingTime <= 60000 is false because of that, I'm not online.

Mardak: `Date.now() - 1000` should result in a number like.. 1465202237419. And a roughly 30 seconds later when I do `Date.now() - 1465202237419` I get 32931.

mancas: My mistake, you're right

mancas:
@Mardak do you wanna review the PR again or may I merge it?

PR #472, opened 2016-05-31
Bug 1276182 - Upgrade marionette and get functional tests working with the new API

Standard8:


Makefile
```diff
@@ -517,8 +517,8 @@ LOOP_SERVER := $(shell echo $${LOOP_SERVER-../loop-server})
 # Either path to the browser, or one of nightly, aurora, beta, firefox.
 TEST_BROWSER := $(shell echo $${TEST_BROWSER-nightly})
 
-ifdef TEST_E10S
-E10S_ARGS = --e10s
+ifndef TEST_E10S
+E10S_ARGS = --disable-e10s
```
dmose, line 522: It'd be good to have a comment about why we've disabled e10s, and info (pointing to a bug, perhaps?) about when we plan to re-enable, given that e10s is going to be the default mode for most users after a while...

test/functional/loopTestDriver.py
```diff
@@ -18,6 +18,10 @@ class LoopTestDriver():
     def setUp(self, marionette):
         self.marionette = marionette
 
+    def set_context(self, context):
+        self.context = context
+        self.marionette.set_context(context)
```
dmose, line 24: It might be nice to file a bug on Marionette requesting get_context() so that this is unnecessary.

test/functional/loopTestDriver.py
```diff
@@ -42,27 +46,37 @@ def wait_for_element_enabled(self, element, timeout=10):
             .until(lambda e: element.is_enabled(),
                    message="Timed out waiting for element to be enabled")
 
-    def wait_for_element_attribute_to_be_false(self, element, attribute, timeout=10):
+    def wait_for_element_property_to_be_false(self, element, property, timeout=10):
+        # XXX We have to switch between get_attribute and get_property here as the
+        # content mode now required get_property for real properties of HTMLElements.
+        # However, in some places (e.g. switch_to_chatbox), we're still operating in
+        # a chrome mode. So we have to use get_attribute for these.
```
dmose, line 54: Probably worth mentioning why we're operating in chrome mode, and point to a bug so that future readers can easily find out when we can get rid of that hack.

test/functional/manifest.ini
```diff
@@ -1,5 +1,7 @@
 [DEFAULT]
-run-if = buildapp == 'browser'
+# app == 'Firefox' is used for running in the Loop repo.
+# buildapp == 'browser' is used for mozilla-central & co.
+run-if = buildapp == 'browser' || app == 'Firefox'
```
dmose, line 4: I'm confused about the buildapp part.  That looks like it would mean that the functional test would be run by default in mozilla-central Mn tests, which, as far as I know, is not what's happening.  Can you explain, and if you think it's worth it, add a comment?

test/functional/loopTestDriver.py
```diff
@@ -42,27 +46,37 @@ def wait_for_element_enabled(self, element, timeout=10):
...
+        def check_property(m):
+            if self.context == "content":
+                return not element.get_property(property)
+
+            return element.get_attribute(property) == "false"
```
dmose, line 61: It's a little surprising that for the property case, we appear to be looking for a falsy value, but for the attribute case, we're looking for only false and not falsy.  I assume that's necessary?  Maybe worth calling out in a comment?

dmose:
r=dmose, with comments addressed as you feel appropriate

test/functional/loopTestDriver.py
Standard8, line 24: Once we sort out the e10s situation, I believe this won't be necessary, so I'm not going to file a bug for that at the moment. I will add a comment.

test/functional/manifest.ini
Standard8, line 4: There's a separate manifest which lists the manifests for Mn tests - the functional test isn't one of those.
The buildapp is existing code, and just says run for the browser tests, as opposed to fennec etc.

test/functional/loopTestDriver.py
Standard8, line 61: "false" is not falsy in python.

PR #474, opened 2016-06-01
Bug 1275104 - static text chat display names

dmose:
Put static displaynames into the UX, based approximately on a mockup/wireframe from [http://i.sevaan.com/063g2Q093g1F/o](http://i.sevaan.com/063g2Q093g1F/o).  This isn't exactly like that mock, but its a solid step preparing for multiple users, so I'd like to land this now so it doesn't bit-rot.

r? @fzzzy @mancas

### DIFF
--- a/shared/css/sidebar.css
+++ b/shared/css/sidebar.css
@@ -259,20 +259,35 @@ html[dir="rtl"] .text-chat-entry.received {
   margin-right: 10px;
 }
 
-.text-chat-entry > p {
-  position: relative;
-  z-index: 10;
+.text-chat-entry p {
   /* Drop the default margins from the 'p' element. */
   margin: 0;
-  padding: .8rem;
+  word-wrap: break-word;
+  line-height: 1.3; /* times the element's font-size */
+}
+
+.speech-bubble {
   /* leave some room for the chat bubble arrow */
   max-width: 70%;
+  position: relative;
+  padding: 0.8rem;
+  z-index: 10;
   border-radius: 15px;
   border: 1px solid #d8d8d8;
   background: #fff;
-  word-wrap: break-word;
+
+  /* part of ensuring that the ordering of the timestamp and the text is
+     different for sent/received status */
   flex: 0 1 auto;
   order: 1;
+}
+
+.text-chat-entry.received .text-chat-entry-displayname {
+  color: #5cccee; /* same as the border */
+}
+
+.text-chat-entry.sent .text-chat-entry-displayname {
+  color: #d8d8d8; /* same as the border */
 }
 
 .text-chat-entry > .context-wrapper {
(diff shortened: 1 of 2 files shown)


shared/css/sidebar.css
```diff
@@ -372,14 +387,26 @@ html[dir="rtl"] .text-chat-entry.received > p:after {
...
+.text-chat-entry-displayname {
+  position: relative;
+
+  /* bolder and smaller are both relative to the parent */
+  font-weight: bolder;
+  font-size: smaller;
+
+  margin-left: -0.2rem; /* outdent this a bit from the main text */
+
+  line-height: 2; /* times the font-size */
```
mancas: nit: Remove new lines between rules or move comments above the rules

shared/css/sidebar.css
```diff
@@ -259,42 +259,57 @@ html[dir="rtl"] .text-chat-entry.received {
   margin-right: 10px;
 }
 
-.text-chat-entry > p {
-  position: relative;
-  z-index: 10;
+.text-chat-entry p {
```
Standard8, line 262: Apologies for the fly-by, but why are we dropping child selectors for the slower performance descendant selectors?

I thought our conventions were to pretty much always to use child selectors, and I can't see a reason to change here.

mancas, line 262: In that case, let's use child selectors everywhere

dmose, line 262: As far as why the change, using child selectors is fragile, particularly in the face of lots of UI change (where we are now with Akita), because you have to correctly edit lots of selectors when the DOM changes much.

Separately from that, though, the logic for this convention mostly grew out of https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Writing_efficient_CSS , and even that document now has a big note at the top linking to [a more modern analysis](http://calendar.perfplanet.com/2011/css-selector-performance-has-changed-for-the-better/), particularly as it relates to selectors, including specifically descendant selectors.  Of interest to Firefox in particular is also https://www.stevesouders.com/blog/2009/03/10/performance-impact-of-css-selectors/#comment-484 -- reading the other comments by Boris make it fairly obvious that it's Boris Zbarsky writing.  In other words, while descendant selectors may have been extremely slow in 2000 when hyatt wrote that doc, that's just not the state of things today.  So in the best case, this is a premature optimization.

All that said, I will concede that inconsistent style sucks, though to some degree, that ship has already sailed in our CSS.  I'll revert these changes if necessary if you still feel strongly about it (do you?), but I just felt the need to grumble about us being held back by an obsolete tradition.  Thank you for tolerating my crankiness.  :-)

Standard8, line 262: > All that said, I will concede that inconsistent style sucks, though to some degree, that ship has already sailed in our CSS. I'll revert these changes if necessary if you still feel strongly about it (do you?), but I just felt the need to grumble about us being held back by an obsolete tradition. Thank you for tolerating my crankiness. :-)

Given the extra pages referenced, I can agree that we don't need to use child selectors everywhere. I'd say that normally we discuss these sort of changes to our procedures/policy before doing them (and there was at least some agreement a while ago to generally switch to child selectors), so please bring this up at the retrospective at least, so that we can all be on the same page.

I would say that I think we do need to continue to bring in the consistency around what CSS applies where - layout within a component to be on that component, layout of a component within a parent component to be applied via the parent. I think that's the biggest thing that'll help us.

Re-reading some of the css here makes it look like we won't really be able to apply that to the text chat if speech-bubble is separated out, but it looks like we probably need to re-architect the rendering this component anyway, so maybe that doesn't matter at the moment.

shared/css/sidebar.css
dmose: Fixed.

shared/css/sidebar.css
dmose, line 262: OK, fair enough.  I'll add something to the retrospective.

PR #475, opened 2016-06-01
Bug 1274177 - Allow deleting records from firebase and detecting thos…

mancas:
…e deletions

### DIFF
--- a/shared/js/dataDriver.js
+++ b/shared/js/dataDriver.js
@@ -94,6 +94,16 @@ loop.DataDriver = function() {
     }
 
     /**
+     * Delete the specified page record.
+     *
+     * @param {String}  pageId   The id to associate with stored page.
+     */
+    deletePage(pageId) {
+      // XXX akita bug 1276095: Verify pageId is a valid firebase id.
+      this.update("page", pageId);
+    }
+
+    /**
      * Send a text chat message by storing in the database.
      *
      * @param {Object} message The message to be stored and received by others.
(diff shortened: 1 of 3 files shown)


shared/js/dataDriver.js
```diff
@@ -413,7 +423,14 @@ loop.DataDriver = function() {
           break;
 
         case "page":
-          dispatchAction = "AddedPage";
+          if (typeof data.value !== "undefined") {
+            dispatchAction = "AddedPage";
```
mancas, line 431: @Mardak shouldn't we include the pageId in the action?

shared/test/dataDriver_test.js
```diff
@@ -80,20 +80,16 @@ describe("loop.DataDriver", () => {
...
 
-    it("should dispatch a AddedPage action for a page record", () => {
-      let record = {
-        title: "cool page",
-        url: "http://example.com"
-      };
-      driver._processRecord("page!00000000", {
-       timestamp: 1234567890123,
-       value: record
-      });
```
mancas, line 92: Just moving the test to suite "#_processRecord.page"

shared/js/dataDriver.js
Mardak, line 431: Probably. Otherwise the UI wouldn't have a good reference to which tile to delete. ;) Similarly, `DeletedPage` should have a `pageId` so that if someone else deletes a page, other open table of contents can update appropriately.

shared/test/dataDriver_test.js
Mardak, line 92: Thanks for fixing this.

PR #476, opened 2016-06-01
Bug 1276945 - Stop the sidebar being created multiple times and more

Standard8:
Also hook up bindings so that it works in the debugger and stop it being remote in non-e10s windows and fix tooltips in non-e10s mode.

### DIFF
--- a/add-on/chrome/modules/LoopSidebar.jsm
+++ b/add-on/chrome/modules/LoopSidebar.jsm
@@ -30,7 +30,7 @@ XPCOMUtils.defineLazyGetter(this, "log", () => {
 var LoopSidebarInternal = {
   /**
    * Creates a sidebar in the given browser window and loads the content based
-   * on the room token
+   * on the room token. It will ensure that the sidebar is only opened once.
    *
    * @param {Object}  window  Browser window where the sidebar will be created
    *
@@ -38,6 +38,11 @@ var LoopSidebarInternal = {
    */
   createSidebar: function(window, token) {
     let ownerDocument = window.gBrowser.ownerDocument;
+
+    if (ownerDocument.getElementById("loop-side-browser")) {
+      return;
+    }
+
     var browser = ownerDocument.getElementById("browser");
 
     let sidebarBrowser = ownerDocument.createElementNS(kNSXUL, "browser");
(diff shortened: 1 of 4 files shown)


add-on/chrome/modules/LoopSidebar.jsm
```diff
@@ -38,6 +38,13 @@ var LoopSidebarInternal = {
    */
   createSidebar: function(window, token) {
     let ownerDocument = window.gBrowser.ownerDocument;
+
+    // We shouldn't ever create more than one sidebar, so just protect against
+    // it here.
+    if (ownerDocument.getElementById("loop-side-browser")) {
+      return;
+    }
+
```
dmose, line 45: Probably worth documenting this behavior in the JSDoc.  I'd suggest also making the MozLoopAPI.createSidebar JSDoc just point to this instead of having a second copy of the doc that needs to be kept in sync.

add-on/chrome/modules/LoopSidebar.jsm
```diff
@@ -47,8 +54,10 @@ var LoopSidebarInternal = {
     sidebarBrowser.setAttribute("frameType", "loop");
     sidebarBrowser.setAttribute("messagemanagergroup", "loop");
     sidebarBrowser.setAttribute("message", "true");
-    sidebarBrowser.setAttribute("remote", "true");
-
+    // If the browser is in non-e10s mode, but we set remote=true, then the sidebar
+    // would run in remote mode regardless.
+    sidebarBrowser.setAttribute("remote", window.gMultiProcessBrowser ? "true" : "false");
+    sidebarBrowser.setAttribute("tooltip", "aHTMLTooltip");
```
dmose, line 57: While the comment is true, it's a little obtuse.  Perhaps phrase it as "ensure that the sidebar only runs in the remote process if e10s is turned on"?

add-on/chrome/skin/shared/loop.css
```diff
@@ -6,6 +6,9 @@
 
 /* Only apply to browser.xul documents */
 @-moz-document url("chrome://browser/content/browser.xul") {
+  browser[frameType="loop"][remote="true"] {
+    -moz-binding: url("chrome://global/content/bindings/remote-browser.xml#remote-browser");
+  }
```
dmose, line 11: Wow; nice catch!

dmose:
Looks good; r=dmose with any appropriate comment tweaks.

PR #477, opened 2016-06-02
Bug 1274179 - Store table of contents pages in Firebase

fzzzy:


shared/js/actions.js
```diff
@@ -145,6 +145,14 @@ loop.shared.actions = (function() {
     }),
 
     /**
+     * Notifies that a tile has been received from the other peer.
+     */
+    ReceivedTile: Action.define("receivedTile", {
+      tile: String
+      // sentTimestamp: String (optional)
```
Mardak: How did you choose these? `tile` seems elsewhere to be `@param {Object} tile  The tile to be stored and received by others.` But we probably actually want multiple fields for url and title at least. Are we expecting to use `sentTimestamp` ?

shared/js/dataDriver.js
```diff
@@ -116,6 +116,15 @@ loop.DataDriver = function() {
       this.update("presence", userId, Object.assign({ isHere }, presence));
     }
 
+    /**
+      * Add tile to table of contents.
+      *
+      * @param {Object} tile  The tile to be stored and received by others.
+      */
+      addTile(tile) {
```
Mardak: This should be placed above `sendTextChatMessage` if we're going with the name `add...` And it should probably be named more explicitly to be a table of content tile as a generic `tile` could mean many things.

shared/js/dataDriver.js
```diff
@@ -417,6 +426,13 @@ loop.DataDriver = function() {
             userId: id
           };
           break;
+
+        case "tocPage":
+          dispatchAction = "ReceivedTile";
+          dispatchExtra = {
+            tile: data.tile
```
Mardak: There is no `data.tile` so this probably isn't doing what you're expecting it to do.

shared/js/actions.js
```diff
@@ -145,6 +145,14 @@ loop.shared.actions = (function() {
     }),
 
     /**
+     * Notifies that a tile has been received from the other peer.
+     */
+    ReceivedTile: Action.define("receivedTile", {
```
Mardak: Name of this action is probably too generic.

shared/test/dataDriver_test.js
```diff
@@ -119,6 +119,31 @@ describe("loop.DataDriver", () => {
     });
   });
 
+  describe("#addTile", () => {
```
Mardak: This should be moved up just before `describe("#sendTextChatMessage", () => {`

shared/js/actions.js
fzzzy: I had intended to switch this to Object as I did in the other place. I'm mildly surprised that even though the types are specified in code they aren't checked.

I wanted to go with Object because I don't think it's clear which fields will be needed here, but I think title and url are a good enough start.

I could see a use for sentTimestamp but I'll remove it.

Mardak: The types are checked when an action is created. Your tests only pass in `String`s.

shared/js/dataDriver.js
fzzzy: Ah, I guess it still worked because tile was in data.value. Whoops.

shared/js/dataDriver.js
```diff
@@ -85,6 +85,18 @@ loop.DataDriver = function() {
     }
 
     /**
+      * Add a page to the room's table of contents.
+      *
+      * @param {String} url       The url of the page.
+      * @param {String} title     The title of the page.
+      * @param {Object} metadata  Additional metadata,
```
Mardak: Any reason to have these as individual parameters instead of a single page object?

shared/js/dataDriver.js
```diff
@@ -85,6 +85,18 @@ loop.DataDriver = function() {
...
+      *                           as provided by PageMetadata.
+      */
+    addPageToRoom(url, title, metadata) {
```
Mardak: The "ToRoom" is making me wonder if we need to have "ToRoom" for all the other APIs. Or if it's unnecessary as all of these are adding data to the room?

shared/js/dataDriver.js
fzzzy: Let me flip this question around: What's the purpose of having a single page object? Will there ever be any other keys at the top level? If not, it seems like a needless extra level of nesting.

Also presumably the url and title will be required but the metadata is always optional.

Mardak: The single object allows the source of the object, e.g., `PageStore`, to add additional fields. I don't know what the structure will be, but I'm not sure if it's useful to have `DataDriver`'s API explicitly encode that. E.g., what if something should move out of `metadata` for some reason? Perhaps the image is important enough, so then we need to expand this API?

And the comment about nesting doesn't quite make sense as a nested object is being created right now for it to be stored in Firebase.

With the single object, `PageStore` can just send over its existing data structure of a page. E.g., some `textChat` objects happen to also have fields for `contentType` and `extraData` that `DataDriver` doesn't even need to be concerned about because there's some abstraction of what a `chat` object needs.

fzzzy: Ok, so you want the method call to take a single parameter but you still want the firebase record to have the values in this object pulled up to the top level of the firebase record? (That's the extra level of nesting I was talking about) No wonder I was so confused.

shared/js/dataDriver.js
fzzzy: We discussed this in irc, and decided "page" is unambiguous enough.

shared/js/dataDriver.js
```diff
@@ -85,15 +85,12 @@ loop.DataDriver = function() {
...
-      *                           as provided by PageMetadata.
-      */
-    addPage(url, title, metadata) {
-      this.update("page", this.makeId(), { title, url, metadata });
+     * Add a page to the room's table of contents.
+     *
+     * @param {Object} page      The page object to add.
+     */
+    addPage(page) {
+      this.update("page", this.makeId(), { title: page.title, url: page.url });
```
Mardak: The page object should be the object that gets dispatched. So just pass in `page` to `update`.

shared/js/actions.js
```diff
@@ -145,6 +145,15 @@ loop.shared.actions = (function() {
     }),
 
     /**
+     * Notifies that a tile has been received from the other peer.
+     */
+    ReceivedAddPage: Action.define("receivedAddPage", {
```
Mardak: Let's just call this `addedPage`

fzzzy: Just to verify, you want this to look like:

```
AddedPage: Action.define("addedPage", {
```

Right?

shared/js/dataDriver.js
fzzzy: So you want it the same way I originally had it before you asked me to change it.

Mardak: My original review comment was about the name `tile` and code positioning. I didn't say to split it into sub objects/fields:

```
This should be placed above sendTextChatMessage if we're going with the name add... And it should probably be named more explicitly to be a table of content tile as a generic tile could mean many things.
```

shared/js/actions.js
Mardak: Yup

shared/js/dataDriver.js
```diff
@@ -417,6 +426,10 @@ loop.DataDriver = function() {
             userId: id
           };
           break;
+
+        case "page":
```
Mardak: nit: move this up before "participant"

PR #444, opened 2016-05-20
Bug 1273336 - Show / Update my name

mancas:

PR #480, opened 2016-06-03
Bug 1274172 - multi-user text chat

dmose:


mozilla-autolander-deprecated:

add-on/panels/js/sidebar.jsx
```diff
@@ -156,6 +156,9 @@ loop.sidebar = (function(mozL10n) {
         dataDriver: dataDriver
       });
 
+      dispatcher.dispatch(
+        new sharedActions.SetOwnDisplayName({ displayName: "Wanda RoomOwner" }));
```
Mardak: Should this be defaulting or just using the "username" pref introduced in https://github.com/mozilla/loop/pull/444/files#diff-705bc24cfa6f5c1421b6e20e28b62660R1495 ?

shared/js/actions.js
```diff
@@ -121,11 +125,20 @@ loop.shared.actions = (function() {
     ReceivedTextChatMessage: Action.define("receivedTextChatMessage", {
       contentType: String,
       message: String,
+      // XXX (optional because we don't use this for tiles.  Should refactor)
+      // name: String,
```
Mardak: This should be `displayName`

shared/js/textChatStore.js
```diff
@@ -148,13 +155,24 @@ loop.store.TextChatStore = (function() {
     },
 
     /**
-     * Handles sending of a chat message.
+     * Handles sending of a chat message.  Adds a name (XXX yuck, this
+     * shouldn't be so side-effecty)
```
Mardak: What do you mean by side-effecty? Because it inserts the `displayName`? We could just have the source that's dispatching `sendTextChatMessage` insert the display name in `textChatView` instead?

It doesn't seem wrong for an action handler to do something like adding a name. Isn't the purpose of a `send` action to cause a side-effect of storing data?

shared/js/textChatStore.js
```diff
@@ -148,13 +155,24 @@ loop.store.TextChatStore = (function() {
...
-      this._dataDriver.sendTextChatMessage(actionData);
+      var messageToSend = {};
+
+      if (actionData.contentType === CHAT_CONTENT_TYPES.TEXT) {
+        // text chat types get the sender's name added
+        Object.assign(messageToSend, actionData,
+          { displayName: this.getStoreState("displayName") }); // XXX no test cov
+      } else {
+        Object.assign(messageToSend, actionData);
+      }
```
Mardak, line 172: This might be a bit simpler:

``` js
let messageToSend = Object.assign({}, actionData);
if (..) {
  Object.assign(messageToSend, { displayName: getStore(..) });
}
```

shared/js/textChatView.jsx
```diff
@@ -87,21 +90,32 @@ loop.shared.views.chat = (function(mozL10n) {
         );
       }
 
+      // Only render messages after they've come back from the server to avoid
+      // dups in a simple way for early product standup.
+      //  XXX akita Probably not the UX we want; should talk to UX folks.
+      if (this.props.type === CHAT_MESSAGE_TYPES.SENT) {
+        return null;
```
Mardak, line 99: Won't this cause all messages to appear blue as they're all RECEIVED now? I suppose that might be part of the XXX comment. Could you file the bug and put the bug number here?

shared/js/textChatView.jsx
```diff
@@ -87,21 +90,32 @@ loop.shared.views.chat = (function(mozL10n) {
...
       if (loop.shared.utils.isDesktop()) {
         linkClickHandler = function(url) {
           loop.request("OpenURL", url);
         };
       }
 
+      var displayName;
+      if (this.props.displayName) {
+        displayName = this.props.displayName;
+      }
```
Mardak: Any reason not to just use `this.props.displayName` below?

shared/js/textChatView.jsx
```diff
@@ -261,6 +275,7 @@ loop.shared.views.chat = (function(mozL10n) {
                 return (
                   <TextChatEntry contentType={entry.contentType}
                                  dispatcher={this.props.dispatcher}
+                                 displayName={entry.displayName ? entry.displayName : undefined}
```
Mardak: I believe `entry.displayName` would be `undefined` in the case that you would have the tertiary result in `undefined` ..?

shared/test/textChatStore_test.js
```diff
@@ -64,13 +70,22 @@ describe("loop.store.TextChatStore", function() {
     });
   });
 
+  describe("#getInitialStoreState", function() {
+    it("should return an object which has no ownUserInfo property", function() {
+      let initialState = store.getInitialStoreState();
+
+      expect(initialState).to.not.have.property("ownUserInfo");
```
Mardak: There's no `ownUserInfo` ever. You're probably wanting to check `displayName` ?

shared/test/textChatStore_test.js
```diff
@@ -149,7 +169,8 @@ describe("loop.store.TextChatStore", function() {
       sinon.assert.calledWithExactly(fakeDataDriver.sendTextChatMessage, messageData);
     });
 
-    it("should add the message to the list", function() {
+    it("should not add the message to the list", function() {
```
Mardak: The test says "not," but the expect below still seems to want a `SENT` message. Probably just have it not have the "not" as it did before as we do add the message to the list but happen to not display it.

shared/test/textChatView_test.js
```diff
@@ -183,7 +182,8 @@ describe("loop.shared.views.TextChatView", function() {
       sinon.assert.notCalled(comp.play);
     });
 
-    it("should show timestamps if there are different senders", function() {
+    // XXX akita need userIds to test for different senders
+    it.skip("should show timestamps if there are different senders", function() {
```
Mardak: This test could still pass I believe. Just update the `SENT` type below to `RECEIVED` simulating multiple other senders.

shared/test/textChatView_test.js
```diff
@@ -203,26 +203,6 @@ describe("loop.shared.views.TextChatView", function() {
           .to.eql(2);
     });
 
-    it("should show timestamps if they are 1 minute apart (SENT)", function() {
```
Mardak, line 206: I have a feeling we'll have SENT messages back soon, so probably best to it.skip and comment for now instead of just deleting.

standalone/content/js/webapp.jsx
```diff
@@ -204,6 +204,9 @@ loop.webapp = (function(_, OT, mozL10n) {
     var participantStore = new loop.store.ParticipantStore(dispatcher);
     var serverConnectionStore = new loop.store.ServerConnectionStore(dispatcher, {});
 
+    dispatcher.dispatch(
+      new sharedActions.SetOwnDisplayName({ displayName: "Bill Smith" }));
```
Mardak: Maybe to spice things up... `"Bill Smith".split("").sort(() => Math.random() - .5).join("")`

E.g.,
B iSllmith
lBilm iSht
lBilS miht
mlBiS tihl
Blil Smthi
BiSllmti h
Bil lmhSit
Btil Smlih
ilBl Smhit
ilB lmSith

Mardak: Eh. Maybe just "Bill Smith " + Math.random().toFixed(1).slice(-1) will be easier to read.

add-on/panels/js/sidebar.jsx
dmose: It should use the username pref, I've got some code for that, I was thinking I'd put it in a spin-off bug.  I'll add XXXakita's with pointers to that bug.

dmose: It's bug 1279042.

shared/js/actions.js
dmose: Good catch; fixed.

shared/js/textChatStore.js
dmose: I was overthinking it.  You're right that the code is fine as is.  I've removed the comment.  I've also added comments and semantic changes to make sure things are ok if the displayName hasn't yet been set.

shared/js/textChatStore.js
dmose, line 172: Good idea.  Done.

shared/js/textChatView.jsx
dmose, line 99: Yes, it does.  

I've filed bug 1279054 for unregressing the speech bubbles and put it in the XXX comment.

shared/js/textChatView.jsx
dmose: Indeed.  Fixed.

shared/js/textChatView.jsx
dmose: Yeah, I think you're right.  Ternary removed.

shared/js/textChatStore.js
dmose: I had to undo the displayName check, as it turns out that the tile code calls this, and it doesn't have any displayName.

shared/test/textChatStore_test.js
dmose: Good catch, this was left over from a previous iteration.

shared/test/textChatStore_test.js
dmose: Quite right; good catch.

shared/test/textChatView_test.js
dmose, line 206: I was operating under the theory "that's what revision control history is for", but ok.

standalone/content/js/webapp.jsx
dmose: Good idea.  I even changed it to "Guest" + ...

shared/test/textChatView_test.js
Mardak, line 206: It's much easier to go back in history/blame by not removing then re-adding.

PR #482, opened 2016-06-03
No Bug. Remove the old now unused conversation.jsx and roomViews.jsx.

Standard8:


mozilla-autolander-deprecated:

PR #483, opened 2016-06-06
Bug 1277089 - Separate room join and refresh

daveccrit:
r? @Standard8


mozilla-autolander-deprecated:

PR #487, opened 2016-06-07
(no bug) Proxy REST requests to the loop server

ianb:
This goes with https://github.com/mozilla-services/loop-server/pull/373 to implement proxying to get around the problem with Firebase

I wasn't sure how to get the value of the loop.server pref in this context.  I hardcoded localhost:5000 for testing.  Obviously this needs to be changed.

There's some notable latency in my experience – we definitely don't want to keep this code!


mozilla-autolander-deprecated:

Mardak:
The change resulting in missing `Access-Control-Allow-Origin: null` header looks to be rolled back on the Firebase side:

https://groups.google.com/d/msg/firebase-talk/nTGbz0q0O1Y/qYqyaWwYHAAJ

Mardak:
This looks unnecessary for now. After the rollback, they switched to `There will be a change to the response headers in the near future.  Requests with the header 'Origin: null' or no origin header will receive a response with 'Access-Control-Allow-Origin: *'.`

PR #490, opened 2016-06-08
Bug 1278677 - Use Embed.ly for temporary metadata collection

mancas:


shared/js/toc.jsx
```diff
@@ -209,7 +211,15 @@ loop.shared.toc = (function(mozL10n) {
...
-      input.value && this.props.handleAddUrlClick(input.value);
+      var url = input.value;
+
+      if (!url) {
+        return;
+      }
+
+      loop.shared.utils.getPageMetadata(url).then(result => {
+        this.props.handleAddUrlClick(result);
+      }).catch(err => console.error(err));
```
fcampo: should we show any error to the user? e.g. invalid URL

shared/js/toc.jsx
```diff
@@ -52,12 +52,14 @@ loop.shared.toc = (function(mozL10n) {
...
+    /**
+     * Adds a new tile to the ToC
+     */
+    addTile: function(metadata) {
       var tiles = this.state.tiles;
       tiles.push({
-        location: url,
-        description: url
+        location: metadata.url,
+        description: metadata.description
```
fcampo, line 63: image adding should be here or a different bug?

shared/js/utils.js
```diff
@@ -812,6 +814,32 @@ if (inChrome) {
     return node;
   }
 
+  function getPageMetadata(url) {
+    return new Promise((resolve, reject) => {
+      let result = { url };
+      let request = new XMLHttpRequest();
+      request.open("GET",
+        `https://api.embedly.com/1/extract?url=${encodeURIComponent(url)}&key=${encodeURIComponent(API_KEY)}`);
+      console.info(request, `https://api.embedly.com/1/extract?url=${encodeURIComponent(url)}&key=${encodeURIComponent(API_KEY)}`);
```
fcampo: is the `console.info` needed here? if it is, can we store the values to avoid double calling?

shared/js/toc.jsx
mancas, line 63: @fcampo It's another bug -> https://bugzilla.mozilla.org/show_bug.cgi?id=1273506

PR #492, opened 2016-06-09
Bug 1277702 - Update presence periodically and check for presence records that are expired

Mardak:


shared/js/participantStore.js
```diff
@@ -39,17 +64,47 @@ loop.store.ParticipantStore = function() {
...
+      }
+
       this._updateParticipantData(actionData.userId, {
         isHere: actionData.isHere,
         localPingTime: Date.now() - actionData.pingedAgo
       });
+
+      // We just got some presence that we might need to expire later.
+      if (this._expireTimer === null) {
+        this._expireTimer = setTimeout(() => this._expirePresence(), PING_TIME);
```
mancas, line 106: Let's add a couple of test to check that the timeout callback is set and called properly

Mardak, line 106: Added checks for timer variable being `null` or not as well as ticking the clock to make sure `_expirePresence` is not called yet or did.

PR #494, opened 2016-06-09
Bug 1273501 - Create thumbnail from screenshot of current state of a URL

chrafuse:
@mancas, @dmose Added the google screenshot url request and response.  Some of the values are not available (description, images, icons).  Produces Base64 image source.  If continuing on this you will need to connect up the variables in the TOC tile correctly as it is giving 'is="null"' in output.  Not sure why.


dmose:
Superceded by https://github.com/mozilla/loop/pull/495, so closing this one.  See the comments there for details.

PR #495, opened 2016-06-10
Bug 1273501 - create thumbnail from screenshot of current state of a url. r=fcampo

dmose:
Since https://github.com/mozilla/loop/pull/494 isn't writable, I forked it.  And, it should be writable by everyone on the team, I think/hope, so this PR will supercede that one.

After the first of these commits, things worked nicely.  

Then I rebased against the latest akita, added the second commit, and now it's broken because the metadata fields in getPageMetadata no longer match the ones that the action and the pageStore expect.  That should be easy to fix up, though.

After that, I think this wants a bit of basic cleanup along with some XXXakita love, as well as possible test tweaks, and we'll have the functionality working for URLs added by the + button.

I've started work based on top of the not-yet-rebased version of this patch over in bug 1274416 to make the initial URL get the metadata used properly as well.

CC @mancas, @crafuse, @fernando, @Standard8, @ianb 

I think anyone in the european timezone who wants to take this bug and run with it should feel free.

### DIFF
--- a/shared/js/actions.js
+++ b/shared/js/actions.js
@@ -160,17 +160,17 @@ loop.shared.actions = (function() {
     /**
      * Used to add a page to the current room
      */
+     // XXX akita: Update actions when metadata extraction bug is landed (Bug 1273497)
     AddPage: Action.define("addPage", {
-      description: String,
-      favicon_url: String,
-      images: Object,
       title: String,
+      thumbnail_img: String,
       url: String
     }),
 
     /**
      * Notifies that a tile has been received from the other peer.
      */
+    // XXX akita rename for readability
     AddedPage: Action.define("addedPage", {
       pageId: String,
       title: String,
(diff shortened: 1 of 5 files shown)


mozilla-autolander-deprecated:

dmose:
(Or, if noone grabs it, we'll find someone US time to dig into tomorrow)...

mancas:
@Standard8 @fcampo Who can do the review?

shared/js/actions.js
```diff
@@ -160,11 +160,10 @@ loop.shared.actions = (function() {
     /**
      * Used to add a page to the current room
      */
+     // XXX akita: Update actions when metadata extraction bug is landed
```
fcampo: please refer to bug number for better tracking

shared/js/pageStore.js
```diff
@@ -65,11 +64,9 @@ loop.store.PageStore = function() {
      */
     addedPage(actionData) {
```
fcampo, line 65: having `addPage` and `addedPage` is a little bit confusing, I'd rather go for `addPage` for the intention of adding a new one, and `pageAdded` for the consequences after adding it, but I understand it's not the scope of this PR, so at the moment, enough with a comment like `// XXX rename for readability`

fcampo, line 65: and same for `deletePage` VS `deletedPage`

PR #499, opened 2016-06-10
Bug 1273507 - Can remove a Tile from the ToC

mancas:


mozilla-autolander-deprecated:

shared/js/pageStore.js
```diff
@@ -91,8 +93,12 @@ loop.store.PageStore = function() {
       let pages = this._storeState.pages;
       pages = pages.filter(page => page.id !== actionData.pageId);
       this.setStoreState({ pages });
+
+      this.dispatchAction(new sharedActions.ShowSnackbar({
+        label: mozL10n.get("snackbar_page_deleted")
+      }));
```
Mardak: So.. this will result in the snackbar showing up if a page has ever been deleted. This happens because we get notified of a deleted page on loading firebase. Perhaps we should check if a page was actually filtered out, so only active users who previously saw the page will see the snackbar.

Mardak: I would just save `pages.length` and check if it's different.

Mardak: Oh and test too with snackbar and without :)

PR #502, opened 2016-06-10
1279551 no chat messaging

daveccrit:
r? @Standard8


mozilla-autolander-deprecated:

PR #503, opened 2016-06-10
Bug 1279551 - Make textChatEnabled true always

daveccrit:
r? @Standard8


Mardak:
`DataDriver` dispatches `DataChannelsAvailable true`. Perhaps we should just have otSdkDriver never dispatch `DataChannelsAvailable false`?

Mardak:
In particular 
https://github.com/mozilla/loop/blob/akita/shared/js/otSdkDriver.js#L826
https://github.com/mozilla/loop/blob/akita/shared/js/otSdkDriver.js#L850

PR #506, opened 2016-06-11
Bug 1279553 - Fix join and leave chat messages

daveccrit:
WIP @Standard8, @Mardak 
My work so far. I will try and finish this tomorrow when I have some time, unless someone else wants to pick it up and finish it.

### DIFF
--- a/locale/en-US/shared.properties
+++ b/locale/en-US/shared.properties
@@ -37,9 +37,10 @@ rooms_room_join_label=Join the conversation
 rooms_room_joined_owner_connected_label2=Your friend is now connected and will be able to see your tabs.
 rooms_room_joined_owner_not_connected_label=Your friend is waiting to browse {{roomURLHostname}} with you.
 
-peer_left_session=Your friend has left.
-peer_unexpected_quit=Your friend has unexpectedly disconnected.
-peer_join_session=Your friend has joined.
+peer_left_session2={{participantName}} has left.
+peer_unexpected_quit2={{participantName}} has unexpectedly disconnected.
+peer_join_session2={{participantName}} has joined.
+peer_no_name=Guest
 
 ## LOCALIZATION NOTE (tos_failure_message): Don't translate {{clientShortname}}
 ## as this will be replaced by clientShortname2.
--- a/shared/js/dispatcher.js
+++ b/shared/js/dispatcher.js
@@ -93,7 +93,7 @@ loop.Dispatcher = (function() {
       this._active = true;
 
       if (this._debug) {
-        console.log("[Dispatcher] Dispatching action", action);
+        console.error("[Dispatcher] Dispatching action", action);
       }
 
       registeredStores.forEach(function(store) {
--- a/shared/js/participantStore.js
+++ b/shared/js/participantStore.js
@@ -14,8 +14,11 @@ loop.store.ParticipantStore = function() {
   const PARTICIPANT_SCHEMA = {
     participantName: "",
     isHere: false,
+    hasNotifiedOfJoin: false,
     localPingTime: null
   };
+
+  var sharedActions = loop.shared.actions;
 
   /**
    * Participant store.
@@ -37,11 +40,9 @@ loop.store.ParticipantStore = function() {
       if (options.updateParticipant) {
         this._currentUserObject = {};
         this.dispatcher.register(this, [
-          "leaveRoom",
           "setOwnDisplayName"
         ]);
       }
-
       this._expireTimer = null;
     },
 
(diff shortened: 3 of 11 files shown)


shared/js/participantStore.js
```diff
@@ -119,6 +120,19 @@ loop.store.ParticipantStore = function() {
         }
       }
 
+      if (!updatedParticipant.hasNotifiedOfJoin &&
+          updatedParticipant.isHere &&
+          updatedParticipant.participantName &&
+          Date.now() - updatedParticipant.localPingTime <= PING_TIME) {
```
Standard8: There's another instance of this calculation in this file, so we should break this out to some sort of helper function.

shared/js/participantStore.js
```diff
@@ -119,6 +120,19 @@ loop.store.ParticipantStore = function() {
         }
       }
 
+      if (!updatedParticipant.hasNotifiedOfJoin &&
+          updatedParticipant.isHere &&
+          updatedParticipant.participantName &&
```
Standard8: Are we always guaranteed a name? Maybe have a fallback just in case someone does something nasty...

Standard8: As in, set up with no name & try to sneak into the room. Unlikely, but possible ;-)

shared/js/participantStore.js
```diff
@@ -138,9 +165,18 @@ loop.store.ParticipantStore = function() {
...
+              localPingTime: participants.localPingTime
+            })) {
             participant.isHere = false;
             expiredAny = true;
+            // Chat has already been notified of this participants join. When
+            // participant presence expires, user is notified of participant
+            // has left
+            if (participant.hasNotifiedOfJoin) {
+              this._sendLeaveNotification(participant.participantName, false);
+              participant.hasNotifiedOfJoin = false;
```
Mardak: Any reason this flag is set outside of the `_sendLeaveNotification` helper?

shared/js/participantStore.js
```diff
@@ -174,6 +210,30 @@ loop.store.ParticipantStore = function() {
...
 
     /**
+     * Handle Participant Join/Leave Message Notification.
+     *
+     * @param participantName
+     * @param hangup
+     */
+    _sendLeaveNotification(participantName, hangup) {
+      this.dispatcher.dispatch(new sharedActions.RemotePeerLeftChat({
+        participantName: participantName,
```
Mardak: Will this end up saying "undefined has left" if there's no name?

locale/en-US/shared.properties
```diff
@@ -38,9 +38,10 @@ rooms_room_join_label=Join the conversation
 rooms_room_joined_owner_connected_label2=Your friend is now connected and will be able to see your tabs.
 rooms_room_joined_owner_not_connected_label=Your friend is waiting to browse {{roomURLHostname}} with you.
 
-peer_left_session=Your friend has left.
-peer_unexpected_quit=Your friend has unexpectedly disconnected.
-peer_join_session=Your friend has joined.
+peer_left_session2={{participantName}} has left.
+peer_unexpected_quit2={{participantName}}  has unexpectedly disconnected.
```
Standard8: nit: double space

shared/js/participantStore.js
```diff
@@ -119,12 +120,38 @@ loop.store.ParticipantStore = function() {
...
 
     /**
+     * Determines if Presence has expired
+     *
+     * @param actionData
+     * @returns {boolean}
+     */
+    _isPresenceExpired(actionData) {
+      if (Date.now() - actionData.localPingTime > PING_TIME) {
+       return true;
```
Standard8: Might as well simplify this down:
`return Date.now() - actionData.localPingTime > PING_TIME;`

shared/js/dispatcher.js
```diff
@@ -92,7 +92,7 @@ loop.Dispatcher = (function() {
       this._active = true;
 
       if (this._debug) {
-        console.log("[Dispatcher] Dispatching action", action);
+        console.log("[Dispatcher] Dispatching action", action, action.name);
```
Standard8: The name is already in the action object, so I don't see the need for this? Or are you trying to compensate for the browser console in e10s mode?

If you're compensating, then I'd suggest putting the action.name first before the action object.

shared/js/participantStore.js
Standard8: We should move these into the function.

shared/js/participantStore.js
Mardak: Are we actually passing in `actionData`? It looks more like a `participant` but why not just the time?

daveccrit:
Not Completed. There seem to be some issues with events not being dispatched correctly. Will need more work or some future commit may fix this issue.

Standard8:

PR #511, opened 2016-06-13
Bug 1274416 - add initial tile w/thumbnail to ToC (WIP)

dmose:
Work in progress patch.  Kinda works, but races and needs to get better metadata.  Oh yeah, and tests.

### DIFF
--- a/shared/js/pageStore.js
+++ b/shared/js/pageStore.js
@@ -38,7 +38,8 @@ loop.store.PageStore = function(mozL10n) {
       "addedPage",
       "deletePage",
       "deletedPage",
-      "setOwnDisplayName"
+      "setOwnDisplayName",
+      "updateRoomInfo"
     ],
 
     /**
@@ -50,9 +51,22 @@ loop.store.PageStore = function(mozL10n) {
 
     /**
      * Handle AddPage action by saving the specific page.
+     *
+     * @note If actionData.url already exists in the store, it will not
+     * be added a second time.
      */
      // XXX akita: Update data saved into Firebase according to the action `AddPage`
     addPage(actionData) {
+
+      // This prevents a race if the link-clicker is opened very
+      // quickly, particularly since the page currently isn't added until
+      // after the metadata fetch completes.  There will still be some kind of
+      // race here even after bug 1281066 lands, but we'l be much less likely to
+      // hit it with any frequency.
+      if (this._hasPageForUrl(actionData.url)) {
+        return;
+      }
+
       this._dataDriver.addPage({
         title: actionData.title,
         thumbnail_img: actionData.thumbnail_img,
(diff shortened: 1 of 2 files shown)


shared/js/pageStore.js
```diff
@@ -100,6 +115,72 @@ loop.store.PageStore = function(mozL10n) {
...
+     * to consider switching to something less strict if it makes sense
+     * from a product standpoint.
+     *
+     * @param {String} url - the thing we're looking for
+     * @return {Boolean}
+     */
+    _hasPageForUrl(url) {
+      let pages = this._storeState.pages;
+      pages = pages.filter(page =>
+        page.url === url
```
mancas, line 139: Please add a XXX akita comment to keep in mind this: `we might want to consider switching to something less strict if it makes sense from a product standpoint.`

shared/js/pageStore.js
```diff
@@ -63,8 +71,15 @@ loop.store.PageStore = function(mozL10n) {
...
     /**
      * Handle AddedPage action by updating the store state.
+     *
+     * @note If actionData.url already exists in the store, it will not
+     * be added a second time.
      */
     addedPage(actionData) {
+      if (this._hasPageForUrl(actionData.url)) {
+        return;
+      }
```
mancas, line 87: Do we really need this second check here? I mean, we don't allow to save duplicate content, so I can't see a case where an `AddedPage` action will throw with duplicate url.

shared/js/pageStore.js
```diff
@@ -100,6 +115,72 @@ loop.store.PageStore = function(mozL10n) {
...
+
+      // XXX do we really care about multiple context urls?  I suspect
+      // we don't.
+      let firstContextUrl = actionData.roomContextUrls[0];
+
+      if (this._hasPageForUrl(firstContextUrl.location)) {
+        return;
+      }
```
mancas, line 170: I'm not sure this check will be useful since the user is creating the room and he can't add content before the creation, right?

shared/js/pageStore.js
dmose, line 139: Done.

shared/js/pageStore.js
dmose, line 87: Because fetching the metadata takes so long, if the link is opened quickly on the guest side, they can race and the link-clicker will see it twice.  I'll add a comment.

shared/js/pageStore.js
dmose, line 170: This mostly prevents us from starting an unnecessary metadata fetch.

dmose, line 170: OH, I see what you mean.  Still, I'd rather have the code behave in a deterministic fashion if something changes or there is an internal bug somewhere.

PR #514, opened 2016-06-13
No bug. Stub localStorage during standaloneMozLoop tests to avoid badly altering localStorage and messing up participants and presence on the standalone UI.

Standard8:


mozilla-autolander-deprecated:

standalone/test/standaloneMozLoop_test.js
```diff
@@ -90,6 +90,26 @@ describe("loop.StandaloneMozLoop", function() {
   });
 
   describe("#rooms.get", function() {
+    var originalLocalStorage;
```
mancas: Why not using `let` instead of `var`?

mancas:
r=me

PR #525, opened 2016-06-18
DO-NOT-MERGE: Greenkeeper eslint 2.13.0 autofix preview

Mardak:
@Standard8, do we want to enable any of these 3 now-autofixable rules? We probably don't want to actually merge this yet as 2.13.0 broke `extends` for us, but we can see what autofixed code looks like.

I needed to disable object-shorthand for karma as that's not es6.

### DIFF
--- a/shared/test/sdk_mock.js
+++ b/shared/test/sdk_mock.js
@@ -19,6 +19,5 @@
   };
 
   window.OT.setLogLevel = function() {};
-
 })(window);
 
--- a/shared/test/sdk_mock.js
+++ b/shared/test/sdk_mock.js
@@ -20,4 +20,3 @@
 
   window.OT.setLogLevel = function() {};
 })(window);
-
--- a/test/karma/.eslintrc
+++ b/test/karma/.eslintrc
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "object-shorthand": [2, "never"]
+  }
+}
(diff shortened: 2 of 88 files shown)


mozilla-autolander-deprecated:

test/karma/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "object-shorthand": [2, "never"]
```
Mardak, line 3: This actually runs object-shorthand and ensures we don't use shorthand (as opposed to just disabling).

shared/test/sdk_mock.js
```diff
@@ -20,4 +20,3 @@
 
   window.OT.setLogLevel = function() {};
 })(window);
-
```
Mardak: Fyi, this was the only trailing space. All the other changes for this rule were for 2 lines -> 1 line.

shared/test/sdk_mock.js
```diff
@@ -19,6 +19,4 @@
   };
 
   window.OT.setLogLevel = function() {};
-
 })(window);
-
```
Mardak, line 24: Fyi, this was the only trailing space. All the other changes for no-multiple-empty-lines rule were for 2 lines -> 1 line.

Mardak:
Probably can skip this for now...

PR #529, opened 2016-06-21
Bug 1280985 - Move presence indicator to the sidebar

mancas:


shared/js/toc.jsx
```diff
@@ -560,6 +563,70 @@ loop.shared.toc = (function(mozL10n) {
...
+      dispatcher: React.PropTypes.instanceOf(loop.Dispatcher).isRequired,
+      participantStore: React.PropTypes.instanceOf(loop.store.ParticipantStore).isRequired,
+      roomState: React.PropTypes.string.isRequired,
+      screen: React.PropTypes.object.isRequired,
+      video: React.PropTypes.object.isRequired
+    },
+
+    getInitialState: function() {
+      return {
+        showRoomControls: this.props.roomState === ROOM_STATES.HAS_PARTICIPANTS
```
mancas: Actually, it doesn't work because the room state is never updated if the webRTC connection isn't established. So, perhaps we should think about a different way to show the media controls. I tried with `activeRoomStore._hasParticipants` and it works for desktop but no for standalone because the room data collected from loop-server in guest don't contain the array of participants

mancas:
Demo using `activeRoomStore._hasParticipants`: [url](https://drive.google.com/file/d/0B4eN5COAQT7_MXJpZ1p2dVJvVDA/view?usp=sharing)

shared/js/toc.jsx
Standard8: I think we should base this on if the participants store is registering participants or not.

The activeRoomStore/WebRTC stuff is now separate, and whilst the loop-server does have an idea of presence, it doesn't get communicated to the standalone.

Hence using Firebase & the participants store is the best thing to do here.

shared/css/sidebar.css
```diff
@@ -103,106 +103,179 @@ body.sidebar {
...
+
+/*------------------------------------*\
     TEXT CHAT
 \*------------------------------------*/
 
 .text-chat-view {
   flex: 0 0 auto;
   /* Leaves enough room for the streams and the media buttons */
-  height: calc(100% - 228px);
+  height: calc(100% - 188px - 40px);
```
fcampo, line 282: can we use the occasion to make chat bigger until presence is detected? there's no reason to save space for avatar if nobody is there

fcampo, line 282: anyway, please add a comment with the origin of those sizes to make it easier to keep it sync'ed

PR #530, opened 2016-06-21
Bug 1281252 - Add better logging for subscribing completing and improve exception logging.

Standard8:


shared/js/otSdkDriver.js
```diff
@@ -502,6 +502,8 @@ loop.OTSdkDriver = (function() {
           break;
         case "Session.networkDisconnected":
         case "Session.forceDisconnected":
+        case "Session.subscribeCompleted":
+        case "Session.screen.subscribeCompleted":
```
dmose, line 506: Why aren't we keeping metrics on (all four of) these?  Because they're logged in some other way?  Probably worth a comment, at the very least.

Standard8, line 506: We do log metrics - these aren't ones that bump the counts but are also valid ones. The default if not listed is to warn and exit the function.

dmose, line 506: Ah, I get it now.  I should have expanded past the bottom the github elider.

shared/js/otSdkDriver.js
```diff
@@ -622,6 +624,7 @@ loop.OTSdkDriver = (function() {
         srcMediaElement: sdkSubscriberVideo
       }));
 
+      this._notifyMetricsEvent("Session.subscribeCompleted");
```
dmose, line 627: It's not obvious why this wants to be logged here rather than at the beginning of the function, since things could blow up before we get here.  Adding a comment about that might be helpful.

shared/js/otSdkDriver.js
```diff
@@ -655,6 +658,7 @@ loop.OTSdkDriver = (function() {
         receiving: true, srcMediaElement: sdkSubscriberVideo
       }));
 
+      this._notifyMetricsEvent("Session.screen.subscribeCompleted");
```
dmose, line 661: This may also want a comment about placement; up to you...

shared/test/otSdkDriver_test.js
```diff
@@ -992,6 +992,23 @@ describe("loop.OTSdkDriver", function() {
...
+          session.trigger("streamCreated", { stream: fakeStream });
+
+          // Called twice due to the VideoDimensionsChanged above.
+          sinon.assert.called(dispatcher.dispatch);
+          sinon.assert.calledWithMatch(dispatcher.dispatch,
+            new sharedActions.ConnectionStatus({
+              event: "Session.subscribeCompleted",
+              state: "receiving",
+              connections: 0,
+              sendStreams: 0,
```
dmose, line 1008: I assume this is 0 because _publishedLocalStream is artificially set?  Probably worth a comment for readability.

dmose, line 1008: And by "this", I think I mean both "connections" and "sendStreams".

shared/test/otSdkDriver_test.js
```diff
@@ -1169,6 +1186,44 @@ describe("loop.OTSdkDriver", function() {
...
+
+            // Called twice due to the VideoDimensionsChanged above.
+            sinon.assert.called(dispatcher.dispatch);
+            sinon.assert.calledWithMatch(dispatcher.dispatch,
+              new sharedActions.ConnectionStatus({
+                event: "Session.screen.subscribeCompleted",
+                state: "receiving",
+                connections: 0,
+                sendStreams: 0,
+                recvStreams: 1
```
dmose, line 1224: If the previous ones deserve comments, this probably does too.

dmose:
Nice work filling in missing test while you were at it!  r=dmose

shared/js/otSdkDriver.js
Standard8, line 627: It was mainly so it was logged near the media connected that happens. It doesn't really matter much though anyway. The bits in the middle are unlikely to blow up unless there's a programming error which would be found before shipping (dispatcher calls are protected).

Standard8, line 627: And even if there was an issue, it'd be better not to log the subscribeCompleted here - as an issue potentially means the subscribe hasn't fully completed anyway.

In any case, I don't think it matters much where this is placed, so long as we log it.

PR #544, opened 2016-06-29
Bug 1279042 - show username from prefs as chat displayName

dmose:
r? @fcampo, @mancas, @daveccrit

### DIFF
--- a/add-on/panels/js/roomToc.jsx
+++ b/add-on/panels/js/roomToc.jsx
@@ -47,9 +47,16 @@ loop.roomToc = (function(mozL10n) {
 
       let pageStore = new loop.store.PageStore(dispatcher, { dataDriver });
 
-      // XXX akita bug 1279042 Use user set name instead of fake name.
-      dispatcher.dispatch(
-        new sharedActions.SetOwnDisplayName({ displayName: "Room Owner" }));
+      // Load the username from storage, or use random guest name
+      loop.request("GetLoopPref", "username")
+        .then(storedName => {
+          let username = storedName ||
+                         "Room Owner"; // XXX akita localize me
+          dispatcher.dispatch(
+            new sharedActions.SetOwnDisplayName({
+              displayName: username
+            }));
+        });
 
       loop.store.StoreMixin.register({
         participantStore,
--- a/add-on/panels/js/panel.jsx
+++ b/add-on/panels/js/panel.jsx
@@ -1481,12 +1481,20 @@ loop.panel = _.extend(loop.panel || {}, (function(_, mozL10n) {
 
       // See nsIPrefBranch
       var PREF_STRING = 32;
-      loop.request("SetLoopPref", "username", newUsername, PREF_STRING).then(() => {
-        this.toggleEditUsername({
-          editionCancelled: true,
-          username: newUsername
+      loop.request("SetLoopPref", "username", newUsername, PREF_STRING)
+        .then(() => {
+          // publish the changes to it takes effect in all the stores
+          this.props.dispatcher.dispatch(
+            new sharedActions.SetOwnDisplayName({
+              displayName: newUsername
+            })
+          );
+          // reset the UI back to normal
+          this.toggleEditUsername({
+            editionCancelled: true,
+            username: newUsername
+          });
         });
-      });
     },
 
     toggleEditUsername: function(extraState) {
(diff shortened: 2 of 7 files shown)


add-on/panels/js/roomToc.jsx
```diff
@@ -47,9 +47,16 @@ loop.roomToc = (function(mozL10n) {
...
       let pageStore = new loop.store.PageStore(dispatcher, { dataDriver });
 
-      // XXX akita bug 1279042 Use user set name instead of fake name.
-      dispatcher.dispatch(
-        new sharedActions.SetOwnDisplayName({ displayName: "Room Owner" }));
+      // Load the username from storage, or use random guest name
+      loop.request("GetLoopPref", "username")
+        .then(storedName => {
+          let username = storedName ||
+                         "Room Owner";
```
mancas: We should add a comment here, to add the default name to locale files later.

shared/js/utils.js
```diff
@@ -172,6 +172,49 @@ if (inChrome) {
...
+      });
+      return;
+    }
+
+    callback(localStorage.getItem(prefName));
+  }
+
+  /**
+   * Used for setting a String preference. It will either use the browser
+   * preferences, or try to get them from localStorage, depending on whether
```
mancas: nit: try to get them from localStorage <- we're setting the preference

shared/js/utils.js
```diff
@@ -172,6 +172,49 @@ if (inChrome) {
...
+  function setStringPreference(prefName, prefValue, callback) {
+    if (loop.shared.utils.isDesktop()) {
+     loop.request("SetLoopPref", prefName, prefValue).then(function() {
+       callback(String(prefValue));
+     });
+
+     return;
+    }
+    //
+    // callback(String(localStorage.getItem(prefName)));
```
mancas: I'm sure we'll need to set the pref from guest as well.

mancas: nit: localStorage.setItem(prefName, prefValue)

shared/test/utils_test.js
```diff
@@ -122,6 +122,115 @@ describe("loop.shared.utils", function() {
...
+    describe("mozLoop not set", function() {
+      beforeEach(function() {
+        sandbox.stub(loop.shared.utils, "isDesktop").returns(false);
+      });
+
+      it("should set the localStorage value", function() {
+        sharedUtils.setStringPreference("test.someStringPref", "sheep", function() {
+          expect(localStorage.getItem("test.someStringPref")).eql("sheep");
+        });
+      });
```
mancas: Not sure why this test is working if the localStorage stuff is commented

shared/js/utils.js
```diff
@@ -172,6 +172,49 @@ if (inChrome) {
...
+  /**
+   * Used for getting a String preference. It will either use the browser
+   * preferences or try to get them from localStorage.
+   *
+   * @param {String}   prefName The name of the preference. Note that mozLoop
+   *                            adds 'loop.' to the start of the string.
+   * @param {Function} callback Invoked with the preference value.  null is
+   *                            returned if no such preference is found.
+   */
+  function getStringPreference(prefName, callback) {
```
fcampo: Not sure we need this function, when we already check from `localStorage` when we use `loop.request("GetLoopPref", pref)` from the standalone side ([standaloneMozLoop](https://github.com/mozilla/loop/blob/master/standalone/content/js/standaloneMozLoop.js#L310-L313))

shared/js/utils.js
```diff
@@ -172,6 +172,49 @@ if (inChrome) {
...
+   * preferences, or try to get them from localStorage, depending on whether
+   * we're on the desktop or not.
+   *
+   * @param {String}   prefName The name of the preference. Note that mozLoop
+   *                            adds 'loop.' to the start of the string.
+   * @param {String}   prefValue
+   * @param {Function} callback Invoked with the preference value for
+   *                            convenience, coerced to String if necessary.
+   */
+  function setStringPreference(prefName, prefValue, callback) {
```
fcampo: [same](https://github.com/mozilla/loop/blob/master/standalone/content/js/standaloneMozLoop.js#L296-L301) here

shared/js/utils.js
dmose: Quite right; I hadn't seen that.  I was following the getBoolPreference pattern, but now I sort of assume that the point of getBoolPreference is less about the encapsulation of preferences, and more about the boolean coercion.  Weird.

add-on/panels/js/roomToc.jsx
dmose: Done.

PR #545, opened 2016-06-29
Bug 1203258 - Hook up the string completeness test to automation. Remove redundant strings.

Standard8:


bin/stringsCompletenessTest.py
```diff
@@ -97,6 +142,7 @@ def check_strings_in_code(l10n_properties, not_found_string_list, **list):
     # directories to look in
     def_code_dirs = [
         "add-on/chrome",
+        "add-on/chrome/modules",
         "add-on/panels/js",
         "add-on/panels/vendor",
```
Mardak, line 147: Do we expect strings from `vendor` files? I would think not, so it could be removed. But if we keep this for `panels`, we probably want to include `shared`/`standalone` equivalent directories too?

bin/stringsCompletenessTest.py
```diff
@@ -64,27 +82,54 @@ def find_l10n_strings(file_path, l10n_properties, not_found_string_list):
             mozL10n\.get\(\"      # mozL10n.get is content code.
             ([a-zA-Z0-9_\.]*?)    # Match any valid string character
             \",.*                 # Only match mozL10n.get("...", ...)
```
Mardak, line 84: Not touched here, but `.get("...")` and `.get("...", ...)` could both be matched with a single regex that allows for either ")" or "," `/mozL10n\.get\("([...]*)"[\),]/` equivalent

locale/en-US/add-on.properties
```diff
@@ -114,45 +114,9 @@ settings_menu_item_feedback=Submit Feedback
...
-## when the call didn't go through.
-call_timeout_notification_text=Your call did not go through.
-
-## LOCALIZATION NOTE (cancel_button):
-## This button is displayed when a call has failed.
-cancel_button=Cancel
 rejoin_button=Rejoin Conversation
 
-cannot_start_call_session_not_ready=Can’t start call, session is not ready.
 network_disconnected=The network connection terminated abruptly.
```
Mardak, line 119: This seems unused. Added https://github.com/mozilla/gecko-dev/commit/1f207797

locale/en-US/add-on.properties
```diff
@@ -170,12 +134,6 @@ legal_text_privacy=Privacy Notice
...
 powered_by_afterLogo=
 
-## LOCALIZATION_NOTE (feedback_rejoin_button): Displayed on the feedback form after
-## a signed-in to signed-in user call.
-feedback_rejoin_button=Rejoin
-## LOCALIZATION NOTE (feedback_report_user_button): Used to report a user in the case of
-## an abusive user.
-feedback_report_user_button=Report User
 feedback_window_heading=How was your conversation?
 feedback_request_button=Leave Feedback
```
Mardak, line 139: While we're removing from add-on, `infobar_screenshare_browser_message2` `login_expired`  `rooms_panel_title` `rooms_signout_alert` additionally seem to be unused. Some added by various land-strings-early commits: https://github.com/mozilla/gecko-dev/commit/6f8791f0 and https://github.com/mozilla/gecko-dev/commit/d424f255

locale/en-US/standalone.properties
```diff
@@ -4,8 +4,6 @@
```
Mardak, line 4: While removing from standalone, `rooms_default_room_name_template` and `vendorShortname` seems to be unused. Added by 1e825246 and 57cffe46

PR #546, opened 2016-06-29
Bug 1281066 - Speed up add new page in Loop table of content

piatra:
Note: I removed the snackbar announcement when a page was added to the TOC. It was a bit more difficult to do in this new implementation, but seeing as how now a page add is instantaneous I think it will not be missed.


piatra:
@mancas || @Mardak please review. whoever gets to it first 😄

shared/test/pageStore_test.js
```diff
@@ -90,6 +90,31 @@ describe("loop.store.PageStore", () => {
...
+    });
+
+    it("should correctly update page matching the id", () => {
+      let updatedPage = {
+        pageId: "fakeId",
+        thumbnail_img: "fakeThumb",
+        title: "fakeTitle"
+      };
+
+      store.updatePage(updatedPage);
```
mancas: Create an action instead of a simple object. As we do in [line 86](https://github.com/mozilla/loop/pull/546/files#diff-fce74e07540ed8189a559e9b22cf3243L86)

shared/test/pageStore_test.js
```diff
@@ -181,8 +220,104 @@ describe("loop.store.PageStore", () => {
...
+            };
+          }
+        };
+      });
+    });
+
+    it("should dispatch UpdatePage", () => {
+      store.fetchPageMetadata(actionData);
+
+      sinon.assert.calledOnce(dispatcher.dispatch);
```
mancas: Please ensure that dispatcher is being called with the correct action.

shared/js/pageStore.js
```diff
@@ -169,25 +209,11 @@ loop.store.PageStore = function(mozL10n) {
...
-              title: firstContextUrl.description,
-              thumbnail_img: thumbnail_img,
-              url: firstContextUrl.location
-            }));
-      });
+      this.dispatchAction(new sharedActions.AddPage({
+            title: firstContextUrl.description,
+            thumbnail_img: firstContextUrl.thumbnail,
+            url: firstContextUrl.location
+          }));
```
Mardak: nit: spacing

shared/test/pageStore_test.js
```diff
@@ -181,8 +220,104 @@ describe("loop.store.PageStore", () => {
...
+    it("should dispatch UpdatePage with results", () => {
+      store.fetchPageMetadata(actionData);
+
+      sinon.assert.calledWithExactly(dispatcher.dispatch,
+        new actions.UpdatePage({
+          pageId: actionData.pageId,
+          thumbnail_img: response.thumbnail_img,
+          title: response.title
+        }));
+    });
```
mancas, line 262: I think we can merge this test and the one above

shared/test/pageStore_test.js
```diff
@@ -181,8 +220,104 @@ describe("loop.store.PageStore", () => {
...
+    it("should dispatch AddPage with correct args", () => {
+      store.updateRoomInfo(actionData);
+
+      sinon.assert.calledWithExactly(dispatcher.dispatch,
+        new actions.AddPage({
+          title: contextUrl.description,
+          thumbnail_img: contextUrl.thumbnail,
+          url: contextUrl.location
+        }));
+    });
```
mancas, line 307: Same here. Merge this test with the one above

shared/js/toc.jsx
```diff
@@ -225,24 +224,32 @@ loop.shared.toc = (function(mozL10n) {
       }
 
       // XXX akita: Code below is just for testing purpose
+      var urlObj;
       var btn = this.refs.addSiteBtn;
       btn.textContent = "Loading...";
       btn.disabled = true;
```
mancas: I think we can remove all the testing code, added because of the waiting for the metadata

shared/js/toc.jsx
```diff
@@ -225,24 +224,32 @@ loop.shared.toc = (function(mozL10n) {
...
-      loop.shared.utils.getPageMetadata(url).then(result => {
+      // Create an URL object in order to use a favicon until the actual
+      // metadata and thumbnail is ready.
+      try {
+        urlObj = new URL(url);
+      } catch (e) {
         this.props.dispatcher.dispatch(new sharedActions.ShowSnackbar({
-          label: mozL10n.get("snackbar_page_added")
+          label: mozL10n.get("snackbar_page_not_added")
         }));
```
mancas, line 229: We can still showing the snackbar. Just dispatch the action from the `AddedPage` method

mancas:
@piatra Let me know when you address all the comments I want to take a quick look after that

shared/test/pageStore_test.js
piatra: That's what the next two test do.

shared/test/pageStore_test.js
piatra, line 262: I think it helps make the tests more easy to debug when the "calledOnce" assertion is separate from the "calledWith" assertion because if the second one fails you can quickly see a green line saying that indeed your stub was called so the problem is with the arguments received.
However I don't feel very strongly about this, I'll update the tests.

shared/js/toc.jsx
piatra, line 229: @mancas Can't do that. Added method is called twice
- when the link is added via + button
- when you open a previous conversation and the links are loaded

piatra:
@mancas updated!

shared/js/toc.jsx
mancas, line 229: Oh, right!

PR #559, opened 2016-07-06
Bug 1283457 - Update username event should spread to the stores

piatra:
Implemented using LoopAPI as described in [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1279042#c3) propagates the event from the panel to the text chat store.


piatra:
@Standard8 @fcampo please review.
Locally functional tests are passing I suspect some task cluster shenanigans but let me know if issue reproduces.

Standard8:
> Locally functional tests are passing I suspect some task cluster shenanigans but let me know if issue reproduces.

That's an intermittent, I've just filed it as [bug 1284805](https://bugzilla.mozilla.org/show_bug.cgi?id=1284805)

add-on/panels/js/roomStore.js
```diff
@@ -124,6 +125,13 @@ loop.store = loop.store || {};
     },
 
     /**
+     * Pass a sharedAction.SetOwnDisplayName from chrome to content space.
+     */
+    setOwnDisplayName: function(actionData) {
+      loop.request("SetOwnDisplayName", actionData);
```
fcampo: as the current flow is `panel.jsx` requesting a change of preferences, and then dispatching the action that `roomStore.js` is receiving, just to request another API call, maybe we can change it for panel requesting this new `SetOwnDisplayName` instead, and the API setting the preference right before broadcasting the message, so no changes on roomStore are needed.
Although I'm not strong on this, it seems, at first glance, less messy for the action/messages back and forth.

shared/js/textChatStore.js
```diff
@@ -49,6 +49,9 @@ loop.store.TextChatStore = (function() {
       }
 
       this._dataDriver = options.dataDriver;
+
+      loop.subscribe("Panel:SetOwnDisplayName",
+                     this.updateOwnDisplayName.bind(this));
```
fcampo, line 54: the `participantStore` should also subscribe to this event, so the presence is updated on the sidebar when the name changes.

add-on/panels/js/roomStore.js
piatra: That makes sense. This patch has no changes to `panel.jsx` the action dispatch was already happening there so I tried to follow that pattern.

piatra:
@fcampo updated.

shared/js/participantStore.js
```diff
@@ -72,6 +75,15 @@ loop.store.ParticipantStore = function() {
     },
 
     /**
+     * Receives action from chrome space.
+     *
+     * @param {sharedActions.SetOwnDisplayName[]} actionData.data
+     */
+    updateOwnDisplayName(actionData) {
+      this.setOwnDisplayName(actionData.data[0]);
```
fcampo, line 83: adding a call to update the room info would trigger a change on the presence indicator for all the guests connected to the room

`this.updateRoomInfo({ userId: this._currentUserId });`

PR #563, opened 2016-07-08
Bug 1285517 - Improve the metrics to provide separation for stream creates for a/v vs screen sharing, and add more information for exceptions.

Standard8:


shared/js/otSdkDriver.js
```diff
@@ -560,8 +563,6 @@ loop.OTSdkDriver = (function() {
      * https://tokbox.com/opentok/libraries/client/js/reference/StreamEvent.html
      */
     _onRemoteStreamCreated: function(event) {
-      this._notifyMetricsEvent("Session.streamCreated");
```
Mardak, line 563: Just making sure that we want to have `Session.streamCreated` be non-screen-only, so comparing this to previous values will be "wrong." (As opposed to keeping this original event's meaning and adding a separate `Session.screen.streamCreated` for just screen.)

Standard8, line 563: The old one wasn't very useful as it actually had two meanings. However, it made me think a bit, and from a stats analysis perspective, it'll be easier if the name changes, and we just drop the old one - hence I've changed `Session.streamCreated` to `Session.av.streamCreated` for better clarity as to when it kicks in.

PR #571, opened 2016-07-15
Bug 1284806 - Investigate switching from Marionette to Selenium+geckodriver for functional tests

Standard8:


test/functional/loopTestDriver.py
```diff
@@ -169,25 +187,28 @@ def local_get_and_verify_room_url(self):
 
     # Assumes the standalone or the conversation window is selected first.
     def check_video(self, selector):
-        video = self.wait_for_element_displayed(By.CSS_SELECTOR, selector, 30)
-        self.wait_for_element_property_to_be_false(video, "paused")
-        self.wait_for_element_property_to_be_false(video, "ended")
+        # XXX This needs re-enabling once selenium is fixed
+        # https://github.com/SeleniumHQ/selenium/issues/2301
```
Standard8, line 191: This issue is the bit that prevents this from landing.

Standard8, line 191: Decided to land this anyway, as what is here is better than what we had previously.

PR #584, opened 2016-08-15
Bug 1292099 - Disable FF Hello from running in FF 49 or later.

Standard8:
Simple change to not start Hello for FF 49 or later (i.e. no button nor anything else) - primarily just for posting to AMO for the beta add-on. We'll do an uninstall itself once we stop running Hello completely.


add-on/chrome/bootstrap.js
```diff
@@ -1383,6 +1383,15 @@ function loadDefaultPrefs() {
  * Called when the add-on is started, e.g. when installed or when Firefox starts.
  */
 function startup(data) {
+  // Anything later than FF 49 we don't support, so don't try to display the
+  // button. Later this will be to uninstall the add-on itself.
+  // Note: this check exists in shutdown() as well.
+  Cu.reportError(Services.appinfo.version);
```
fzzzy: Is reportError the right thing to use here? Should there be a more detailed message here so people know where this message comes from?

